PR #87, opened 2018-11-10
Fix: #76 User can save correct number now

### LINKED ISSUE #76: Cannot change Mobile Number
**Summary:**

There is option to change the mobile number but it gives an error.

**Steps to reproduce:**

Dashboard -> Profile -> Edit Profile -> Change Mobile Number.

**Expected behavior:**

I would be able to change the Mobile Number without any error.

**Observed behavior:**

When tried to change the email Id, error pops up.

Android Emulator Pixel 2 (Marshmallow)

letelete:
Fixes #76 

Please Add Screenshots If any UI changes.

ezgif-4-07b2895af235

Please Add Summary of what changes you have made. (Optional)

The bug occurred because of country code which was added to the input. For now, only for input case, we format mobile number by excluding country code, so always if the number is correct it will be saved.

I've also added the length limit to the input which equals 15 characters (International standard).


letelete:
I've tested it with various countries (India including) and everything works fine.

ankurs287:
Good Job, again!! @letelete

PR #125, opened 2018-12-07
Feat: Redesign Payments Tab

### LINKED ISSUE #122: Redesign: Prepare Payments for the redesign
**NOTICE** 

**Remember to checkout and send a pull request to the [redesign](https://github.com/openMF/mobile-wallet/tree/redesign)  branch.**

---

We are working on the app redesign. The prototype is done and can be found here: [show prototype](https://xd.adobe.com/view/8581e34d-8b91-49ec-7c21-000d98c8f18b-f3b0/?hints=off), and source files here: [show repository](https://github.com/letelete/mifos-pay-app-design). We will be working on the redesign implementation now.

This issue is about **preparing** a layout for the payments section redesign and new features implementation. You will need to create a tab layout for the payments section (Leave tabs empty for now), prepare fragments for the sections and check if you can easily move around these tabs.

**NOTICE** This issue is NOT about redesigning a layout or implementing new features, and only about preparing Payments section for that.

That's what we want to achieve:

image

m-sameer:
Fixes #122 

- Redesign Payments Tab and transfer fragment has been moved to the Finance tab
- I have used a dummy TextView in the ViewPager

Checks:

- [x] Apply the `AndroidStyle.xml` style template to your code in Android Studio

- [x] If you have multiple commits please combine them into one commit by squashing them

### DIFF
--- a/mifospay/src/main/java/org/mifos/mobilewallet/mifospay/home/adapter/PaymentsAdapter.java
+++ b/mifospay/src/main/java/org/mifos/mobilewallet/mifospay/home/adapter/PaymentsAdapter.java
@@ -1,0 +1,42 @@
+package org.mifos.mobilewallet.mifospay.home.adapter;
+
+import android.support.annotation.Nullable;
+import android.support.v4.app.Fragment;
+import android.support.v4.app.FragmentManager;
+import android.support.v4.app.FragmentStatePagerAdapter;
+
+import org.mifos.mobilewallet.mifospay.payments.ui.HistoryFragment;
+import org.mifos.mobilewallet.mifospay.payments.ui.InvoicesFragment;
+import org.mifos.mobilewallet.mifospay.payments.ui.RequestFragment;
+import org.mifos.mobilewallet.mifospay.payments.ui.SendFragment;
+
+public class PaymentsAdapter extends FragmentStatePagerAdapter {
+    private final Fragment[] mFragments = {
+            new SendFragment(),
+            new RequestFragment(),
+            new HistoryFragment(),
+            new InvoicesFragment()
+    };
+    private final String[] mTitles;
+
+    public PaymentsAdapter(FragmentManager fm, String[] titles) {
+        super(fm);
+        mTitles = titles;
+    }
+
+    @Override
+    public Fragment getItem(int position) {
+        return mFragments[position];
+    }
+
+    @Override
+    public int getCount() {
+        return mFragments.length;
+    }
+
+    @Nullable
+    @Override
+    public CharSequence getPageTitle(int position) {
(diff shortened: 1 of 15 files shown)


luckyman20:
@m-sameer Please see the prototypes carefully, you have created blank screens whereas you need to create contents in it too.

ankurs287:
@m-sameer There should be appropriate placeholders when there is no history/invoices/etc.

letelete:
It's my bad I didn't mention that in the issue description but would be great if you could add placeholders for the `HISTORY` and `INVOICES` tabs since these might be empty.

(Just like in the Transactions History)

72f8a9b5-4296-4ee4-8110-8a507251d4d7

m-sameer:
@ankurs287 I have added empty states to history and invoice tab but there are no empty states for send and request so I have added the first **TextView** in the screen.

letelete:
Btw. I think we could improve code quality  and avoid code repeat by creating `setupStateView(int res, String title, String subtitle)` method in the BaseFragment/Activity

m-sameer:
> Btw. I think we could improve code quality and avoid code repeat by creating `setupStateView(int res, String title, String subtitle)` method in the BaseFragment/Activity

@letelete The first parameter of the method, I think, must be the resource id of the drawable resource. We cannot just have the drawable resource, title, and subtitle as the parameter we will also going to need the **ImageView** and two **TextView**s as the parameter. That will be 6 parameters and 6 parameters for a single method are too much as Java doesn't support default parameters so every time all the 6 parameters should be provided which is not much liked by programmers in general.

letelete:
Oh yeah, you're right about drawable. I was just quickly checking the class so didn't notice the drawable in the argument.

We don't need views as parameters, just take a look at how toolbar is implemented. It's going to be 3 arguments in the method.

letelete:
Btw since Base is an activity/fragment we can still handle getResources.getDrawable(**res**) easily and improve readability by changing the method to the setupStateView(**int res**, String title, String subtitle)

letelete:
So the example call could look like this: 

```
private void showErrorState() {
        setupErrorState(R.id.error_drawable, getString(R.string.title),
                getString(R.string.subtitle));
}
```

We're can save so many redundant code repeat by this solution

m-sameer:
> We don't need views as parameters, just take a look at how toolbar is implemented. It's going to be 3 arguments in the method.

I think we do need views as parameters, toolbar works because we set the toolbar title in **BaseActivity** by calling `setTitle`. We cannot do the same in case of these views. We will need the view so set its image/text.

letelete:
I'm not sure about it, but if its true, we can make use of custom views in this case and it still will reduce code repeats

PR #156, opened 2018-12-20
Added setError attribute to amount if ScanQrCode is pressed

### LINKED ISSUE #155: App crashes on scanning QR code if amount is empty
**Summary:**
App crashes on scanning QR code if amount field is empty.

**Steps to reproduce:**
How can we reproduce again the issue?
1) Navigate to transfer fragment and without entering amount click Scan QR code. 
2) Scan any code 
3) App crashes (BUG)

**Expected behavior:**
The user should be alerted to first enter the amount

OnePlus 5 
Android Oreo 8.1

codepoet2017390:
Fixes #155 

**SCREENSHOT**

screenshot_20181220-130643

### DIFF
--- a/mifospay/src/main/java/org/mifos/mobilewallet/mifospay/home/ui/TransferFragment.java
+++ b/mifospay/src/main/java/org/mifos/mobilewallet/mifospay/home/ui/TransferFragment.java
@@ -149,8 +149,17 @@ public class TransferFragment extends BaseFragment implements HomeContract.Trans
         String eamount = etAmount.getText().toString().trim();
         String mobileNumber = mEtMobileNumber.getText().toString().trim().replaceAll("\\s+", "");
         if (eamount.equals("") || (externalId.equals("") && mobileNumber.equals(""))) {
-            Toast.makeText(getActivity(),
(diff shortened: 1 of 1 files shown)


luckyman20:
Good work! But we need one for VPA too. Can you please add that too as VPA can never be empty?

codepoet2017390:
@luckyman20  Yes i already had that in consideration. The Scan QR code directly adds its results to the VPA and if u press proceed with VPA empty it alerts via Toast message

luckyman20:
So in that case we can remove toast for VPA and add this type of error in the VPA too.

codepoet2017390:
Should i open another issue for it?as per the guidelines

luckyman20:
You can always edit the existing one and then do the required changes.

codepoet2017390:
@luckyman20 these are to separate changes so do i still need to stash the commits

codepoet2017390:
@luckyman20  Actually the App crashes because we are passing amount field to a further activity that is called on Scanning QR my main aim was to prevent crash . I will update my PR with all the errors fixed

luckyman20:
I see that we aren't using amount which is being sent, so you can simply remove that from Intent.

PR #268, opened 2019-02-03
refactor : Replaced String with StringBuilder

### LINKED ISSUE #259: Replace String with StringBuilder
**Summary:**
As we are using String for appending [here](https://github.com/openMF/mobile-wallet/blob/master/mifospay/src/main/java/org/mifos/mobilewallet/mifospay/utils/DebugUtil.java#L12) in `DebugUtil.java` and [here](https://github.com/openMF/mobile-wallet/blob/master/mifospay/src/main/java/org/mifos/mobilewallet/mifospay/kyc/presenter/KYCLevel2Presenter.java#L66) in `KYCLevel2Presenter.java`. It is better to use StringBuilder as it is mutable. A string is immutable and when you invoke a method on String it creates a new object and returns it. My idea is to place String with StringBuilder.
Here is the [link](https://stackoverflow.com/questions/5234147/why-stringbuilder-when-there-is-string) for more explanation.
This type of issues will throw warnings which comes under Java --> Performance warnings.

RajaVamsi11:
Fixes #259 

Replaced String with StringBuilder in two java files for better performance.

### DIFF
--- a/mifospay/src/main/java/org/mifos/mobilewallet/mifospay/kyc/presenter/KYCLevel2Presenter.java
+++ b/mifospay/src/main/java/org/mifos/mobilewallet/mifospay/kyc/presenter/KYCLevel2Presenter.java
@@ -63,10 +63,11 @@ public class KYCLevel2Presenter implements KYCContract.KYCLevel2Presenter {
                 intent.putExtra(Intent.EXTRA_MIME_TYPES, mimeTypes);
             }
         } else {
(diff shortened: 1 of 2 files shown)


RajaVamsi11:
@naman14 @ankurs287 Please review this PR.

luckyman20:
@RajaVamsi11, @codepoet2017390  is right, no major improvement in time complexity.

PR #294, opened 2019-02-12
fix : App crashing on first build

### LINKED ISSUE #293: App crashes on first build
**Summary:**

After importing the project in Android studio, while running for the first time the app crashes throwing Null pointer expection in `MifosPayInstanceIdService.java` and works fine in next tries.

 **Steps to reproduce:**

- Build the project

- Run the app for first time ( Make sure your Internet is on )

**Expected behavior:**

App should not crash.

**Observed behavior:**

While running for the first time the app crashes throwing this exception
`java.lang.NullPointerException
        at org.mifos.mobilewallet.mifospay.data.firebase.api.services.MifosPayInstanceIDService.onTokenRefresh(MifosPayInstanceIDService.java:52)` 
and it runs without crash in next time. 

Device Model : Asus zenfone 5
Android version : 4.4.0

**Screenshots:**

**On first build**
screenshot_2019-02-12-15-37-22
**On next try**
screenshot_2019-02-12-15-37-35

RajaVamsi11:
Fixes #293 

### DIFF
--- a/mifospay/src/main/java/org/mifos/mobilewallet/mifospay/data/firebase/api/services/MifosPayInstanceIDService.java
+++ b/mifospay/src/main/java/org/mifos/mobilewallet/mifospay/data/firebase/api/services/MifosPayInstanceIDService.java
@@ -21,6 +21,7 @@ import android.util.Log;
 import com.google.firebase.iid.FirebaseInstanceId;
 import com.google.firebase.iid.FirebaseInstanceIdService;
 
+import org.mifos.mobilewallet.mifospay.MifosPayApp;
 import org.mifos.mobilewallet.mifospay.data.local.PreferencesHelper;
 
 import javax.inject.Inject;
@@ -32,9 +33,12 @@ import javax.inject.Inject;
 public class MifosPayInstanceIDService extends FirebaseInstanceIdService {
 
     private static final String TAG = "MyFirebaseIIDService";
-
     @Inject
     PreferencesHelper mPreferencesHelper;
+
+    public MifosPayInstanceIDService() {
+        MifosPayApp.get(MifosPayApp.getContext()).component().inject(this);
+    }
 
     /**
      * Called if InstanceID token is updated. This may occur if the security of
--- a/mifospay/src/main/java/org/mifos/mobilewallet/mifospay/injection/component/ApplicationComponent.java
+++ b/mifospay/src/main/java/org/mifos/mobilewallet/mifospay/injection/component/ApplicationComponent.java
@@ -6,6 +6,7 @@ import android.content.Context;
 import org.mifos.mobilewallet.core.base.UseCaseHandler;
 import org.mifos.mobilewallet.core.data.fineract.api.FineractApiManager;
 import org.mifos.mobilewallet.core.data.fineract.repository.FineractRepository;
+import org.mifos.mobilewallet.mifospay.data.firebase.api.services.MifosPayInstanceIDService;
 import org.mifos.mobilewallet.mifospay.data.local.LocalRepository;
 import org.mifos.mobilewallet.mifospay.data.local.PreferencesHelper;
 import org.mifos.mobilewallet.mifospay.injection.ApplicationContext;
@@ -36,4 +37,5 @@ public interface ApplicationComponent {
     LocalRepository localRepository();
 
 
+    void inject(MifosPayInstanceIDService mifosPayInstanceIDService);
 }


RajaVamsi11:
@luckyman20 @ankurs287 Can you please review this PR?

mifospay/src/main/java/org/mifos/mobilewallet/mifospay/data/firebase/api/services/MifosPayInstanceIDService.java
```diff
@@ -44,23 +44,25 @@
...
-        // Saving reg id to shared preferences
-        mPreferencesHelper.setFirebaseRegId(refreshedToken);
+            // Saving reg id to shared preferences
+            mPreferencesHelper.setFirebaseRegId(refreshedToken);
 
-        // If you want to send messages to this application instance or
-        // manage this apps subscriptions on the server side, send the
-        // Instance ID token to your app server.
-        sendRegistrationToServer(refreshedToken);
+            // If you want to send messages to this application instance or
```
luckyman20: Remove unnecessary changes.

RajaVamsi11:
@luckyman20 Done.

mifospay/src/main/java/org/mifos/mobilewallet/mifospay/data/firebase/api/services/MifosPayInstanceIDService.java
RajaVamsi11: ok

mifospay/src/main/java/org/mifos/mobilewallet/mifospay/data/firebase/api/services/MifosPayInstanceIDService.java
```diff
@@ -44,6 +44,7 @@
     // [START refresh_token]
     @Override
     public void onTokenRefresh() {
+        if(mPreferencesHelper != null){
```
luckyman20:
```suggestion
        if(mPreferencesHelper != null) {
```

RajaVamsi11:
@luckyman20 Updated.

mifospay/src/main/java/org/mifos/mobilewallet/mifospay/data/firebase/api/services/MifosPayInstanceIDService.java
```diff
@@ -44,6 +44,7 @@
     // [START refresh_token]
     @Override
     public void onTokenRefresh() {
+        if(mPreferencesHelper != null) {
```
luckyman20: What if it will be `null`? Please explain.

RajaVamsi11: Simply it will throw Null Pointer exception when it is null. When the PreferenceHelper object is initialized it contains nothing and here we are calling setFireBaseRegId(refreshedToken) on that object. As it contains nothing, it is throwing Null pointer exception.

RajaVamsi11:
@luckyman20 Can I change documentation according to the newly documented files? Adding space is breaking the fix. If we change documentation accordingly it will be fine. What do you say?

luckyman20:
Looks good to me. @RajaVamsi11 For getting the indentation right, first select all the code (`Ctrl + A`) and then format the code (`Ctrl + Alt + L`).

RajaVamsi11:
@luckyman20 I think the code is properly intended to the changes I made. I checked it using (ctrl+alt+L).

luckyman20:
Yeah now it is properly intended, I was talking for future changes.

RajaVamsi11:
@luckyman20  Ok, thank you. I think it is safe to merge the PR now. What do you say?

mifospay/src/main/java/org/mifos/mobilewallet/mifospay/data/firebase/api/services/MifosPayInstanceIDService.java
naman14: This is incorrect.

The app crashes on first time since it is null and it doesn't crashes next time since `onTokenRefresh` is not even called. (Its only called when the token actually refreshes e.g. new install, app update etc).

See the comment in PR for more details

naman14:
@RajaVamsi11 This is not a fix
You are just checking if `mPreferencesHelper` is not null.

The actual issue here is that the dependency is not being injected anywhere so `mPreferencesHelper` will always be null and not just first time, its just that onTokenRefresh is only called first time

You need to inject the `PreferencesHelper` in constructor

mifospay/src/main/java/org/mifos/mobilewallet/mifospay/data/firebase/api/services/MifosPayInstanceIDService.java
```diff
@@ -44,17 +44,19 @@
...
-        Log.d(TAG, "Refreshed token: " + refreshedToken);
-
-        // Saving reg id to shared preferences
-        mPreferencesHelper.setFirebaseRegId(refreshedToken);
-
-        // If you want to send messages to this application instance or
-        // manage this apps subscriptions on the server side, send the
-        // Instance ID token to your app server.
-        sendRegistrationToServer(refreshedToken);
+        if (mPreferencesHelper != null) {
```
naman14: Remove this check and inject `mPreferencesHelper` properly

RajaVamsi11: Done

RajaVamsi11:
@naman14  Updated according to the changes suggested. Can you please review the PR.

mifospay/src/main/java/org/mifos/mobilewallet/mifospay/data/firebase/api/services/MifosPayInstanceIDService.java
```diff
@@ -32,9 +32,12 @@
 public class MifosPayInstanceIDService extends FirebaseInstanceIdService {
 
     private static final String TAG = "MyFirebaseIIDService";
+    private final PreferencesHelper mPreferencesHelper;
 
     @Inject
-    PreferencesHelper mPreferencesHelper;
+    public MifosPayInstanceIDService(PreferencesHelper mPreferencesHelper) {
```
naman14: `FirebaseInstanceIdService` needs a zero args constructor so you cant provide a constructor that has `PreferencesHelper` argument. (It will only call the zero arg constructor, will also be crashing in your latest change)

Instead do it like this - 
```
    @Inject
    PreferencesHelper mPreferencesHelper;

    public MifosPayInstanceIDService() {
        MifosPayApp.get(MifosPayApp.getContext()).component().inject(this);
    }
```

And add this method in `ApplicationComponent`
```
    void inject(MifosPayInstanceIDService mifosPayInstanceIDService);
```

RajaVamsi11:
@naman14 Updated.

PR #379, opened 2019-02-28
[REDESIGN -> DEV] Finalise redesign, sync to dev

naman14:


naman14:
@letelete @ankurs287 @luckyman20 We need to rebase `redesign` with `dev` and fix conflicts.
Since I don't have much context about the redesign, it would be helpful if you guys can take this forward and sync dev with redesign asap and make it in a mergeable state

PR #417, opened 2019-03-05
feat : Added Offline Profile Pic Storage and Pic view

### LINKED ISSUE #365: User's profile image is not saved in a database.
**Summary:**

UI should be saved locally for the user.

**Steps to reproduce:**

Upload user image in edit profile tab.

**Expected behavior:**

Locally saving image will lead to just retrieving the save location and not selecting the image again.
Even the removeProfileImage() method can be applied.
~~~ java
    @Override
    public void removeProfileImage() {
        // TODO: Remove image from database
    }
~~~

**Observed behavior:**

I didn't see any local storage of user's profile image.

Redmi Note 3

Vipinsoni38:
Fixes #365 

Currently, if a user adds his profile pic it just gets added in one section and doesn't get saved ( in Server )
Implemented a function to save that extracted Bitmap locally and fetch it every time we need to use it 

also added the profile pic viewer as once the user has added his profile pic he can view it in an animated dialog box as shown in GIF

### DIFF
--- a/mifospay/src/main/java/org/mifos/mobilewallet/mifospay/editprofile/ui/EditProfileActivity.java
+++ b/mifospay/src/main/java/org/mifos/mobilewallet/mifospay/editprofile/ui/EditProfileActivity.java
@@ -7,6 +7,7 @@ import android.graphics.Bitmap;
 import android.net.Uri;
 import android.os.Build;
 import android.os.Bundle;
+import android.provider.MediaStore;
 import android.support.annotation.NonNull;
 import android.support.design.widget.BottomSheetDialog;
 import android.support.v4.content.ContextCompat;
@@ -21,12 +22,14 @@ import android.widget.TextView;
 import com.hbb20.CountryCodePicker;
 import com.yalantis.ucrop.UCrop;
 
+import java.io.IOException;
 import org.mifos.mobilewallet.mifospay.R;
 import org.mifos.mobilewallet.mifospay.base.BaseActivity;
 import org.mifos.mobilewallet.mifospay.editprofile.EditProfileContract;
 import org.mifos.mobilewallet.mifospay.editprofile.presenter.EditProfilePresenter;
 import org.mifos.mobilewallet.mifospay.utils.AnimationUtil;
 import org.mifos.mobilewallet.mifospay.utils.Constants;
+import org.mifos.mobilewallet.mifospay.utils.ImageInputOutput;
 import org.mifos.mobilewallet.mifospay.utils.TextDrawable;
 import org.mifos.mobilewallet.mifospay.utils.Toaster;
 import org.mifos.mobilewallet.mifospay.utils.ValidateUtil;
@@ -110,6 +113,7 @@ public class EditProfileActivity extends BaseActivity implements
     private String email;
     private String mobile;
     private BottomSheetDialog bottomSheetDialog;
+    private ImageInputOutput imageInputOutput;
 
     class BottomSheetViews {
         @OnClick(R.id.ll_change_profile_image_dialog_row)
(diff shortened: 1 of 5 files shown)


shaliniv16:
@Vipinsoni38 Why did you open a PR regarding this issue #365? @nikhil-sachdeva has already said that he
would send a PR regarding this issue.

Vipinsoni38:
@shaliniv16  I started my work before his comment !
and that issue was a bit old
so started working on it

mifospay/src/main/java/org/mifos/mobilewallet/mifospay/editprofile/ui/EditProfileActivity.java
```diff
@@ -138,7 +143,10 @@ protected void onCreate(Bundle savedInstanceState) {
 
         showProgressDialog(Constants.PLEASE_WAIT);
         mEditProfilePresenter.fetchUserDetails();
-
+        Bitmap bitmap = load();
```
droidchef: There are multiple problems with this line.
1. The File is being read on the main thread.
2. The File is being read inside the activity which is wrong because business logic needs to live in the `Presenter`.

mifospay/src/main/java/org/mifos/mobilewallet/mifospay/editprofile/ui/EditProfileActivity.java
```diff
@@ -363,7 +371,10 @@ protected void onActivityResult(int requestCode, int resultCode, Intent data) {
                     startCrop(selectedImageUri);
                 }
             } else if (requestCode == UCrop.REQUEST_CROP) {
-                handleCropResult(data);
+                try {
+                    handleCropResult(data);
+                } catch (IOException e) {
```
droidchef: Exception not handled correctly.

mifospay/src/main/java/org/mifos/mobilewallet/mifospay/editprofile/ui/EditProfileActivity.java
droidchef: * File being read on main thread.
* Business logic should not live in the activity but go into presenter.

mifospay/src/main/java/org/mifos/mobilewallet/mifospay/editprofile/ui/EditProfileActivity.java
```diff
@@ -509,4 +523,45 @@ public void setImage(String fullName) {
         mIvUserImage.setImageDrawable(drawable);
         hideProgressDialog();
     }
+    public void savebitmap(Bitmap bitmapImage) {
```
droidchef: * Method doesn't need to be `public`.

* Do File I/O on a separate thread not on the main thread.

mifospay/src/main/java/org/mifos/mobilewallet/mifospay/editprofile/ui/EditProfileActivity.java
```diff
@@ -509,4 +523,45 @@ public void setImage(String fullName) {
...
+        try {
+            inputStream = new FileInputStream(createFile());
+            return BitmapFactory.decodeStream(inputStream);
+        } catch (Exception e) {
+        } finally {
+            try {
+                if (inputStream != null) {
+                    inputStream.close();
+                }
+            } catch (IOException e) {
```
droidchef: Unhandled exceptions

droidchef (changes requested):
Lot's of problems with the PR. I have left some comments, they need to addressed before this can be merged.

Please pay special attention to the architecture of the app.

mifospay/src/main/java/org/mifos/mobilewallet/mifospay/editprofile/ui/EditProfileActivity.java
Vipinsoni38: So should i simply log the Error and proceed

mifospay/src/main/java/org/mifos/mobilewallet/mifospay/editprofile/ui/EditProfileActivity.java
Vipinsoni38: Done !

mifospay/src/main/java/org/mifos/mobilewallet/mifospay/editprofile/ui/EditProfileActivity.java
Vipinsoni38: Created ImageInputOutput.java class  👍

mifospay/src/main/java/org/mifos/mobilewallet/mifospay/editprofile/ui/EditProfileActivity.java
Vipinsoni38: Resolved :)

Vipinsoni38:
@droidchef  Resolved all the suggestions 
created separate class keeping in mind the architecture of the app

PR #424, opened 2019-03-07
fix : Added trim() method to all EditText objects in the whole project

### LINKED ISSUE #423: Add trim() method to all EditText objects in the project
**Summary:**

Add .trim() method to all EditText objects in the whole project. This method returns a copy of the string, with leading and trailing whitespace omitted. The Unicode value of space character is ‘\u0020’. The trim() method in java checks this Unicode value before and after the string, if it exists then removes the spaces and returns the omitted string.

**Example**
For example, In the edit text field, we wrote something like "  issue    " and submitted. When there is no trim method applied on that string, the string will be stored as "  issue    " with spaces included at the first and last of the string. But when a trim method is applied the string is stored as "issue" with spaces removed from before and after the string. This is how it looks like after applying:

                       @BindView(R.id.et_username)
                       EditText etUsername;
                       etUserName.getText().toString().trim()

This must be implemented in whole project. You can refer [here](https://www.geeksforgeeks.org/java-string-trim-method-example/) and [here](https://www.tutorialspoint.com/java/java_string_trim.htm) for more information.

RajaVamsi11:
Fixes #423 

### DIFF
--- a/mifospay/src/main/java/org/mifos/mobilewallet/mifospay/bank/ui/LinkBankAccountActivity.java
+++ b/mifospay/src/main/java/org/mifos/mobilewallet/mifospay/bank/ui/LinkBankAccountActivity.java
@@ -127,7 +127,7 @@ public class LinkBankAccountActivity extends BaseActivity implements
             public void onTextChanged(CharSequence s, int start, int before, int count) {
 //                Log.d("qxz", "onTextChanged: " + s.toString());
 //                mOtherBankAdapter.getFilter().filter(mEtSearchBank.getText().toString());
-                filter(mEtSearchBank.getText().toString());
+                filter(mEtSearchBank.getText().toString().trim());
             }
 
             @Override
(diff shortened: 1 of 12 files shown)


RajaVamsi11:
@naman14 @droidchef Can you please review this PR.

mifospay/src/main/java/org/mifos/mobilewallet/mifospay/bank/ui/LinkBankAccountActivity.java
```diff
@@ -127,7 +127,7 @@ public void beforeTextChanged(CharSequence s, int start, int count, int after) {
             public void onTextChanged(CharSequence s, int start, int before, int count) {
 //                Log.d("qxz", "onTextChanged: " + s.toString());
 //                mOtherBankAdapter.getFilter().filter(mEtSearchBank.getText().toString());
-                filter(mEtSearchBank.getText().toString());
+                filter(mEtSearchBank.getText().toString().trim());
```
luckyman20: Bank name can have spaces.

luckyman20 (commented):
@RajaVamsi11 Remove `trim()` method from all the places where we are searching something.

rchtgpt:
Thanks for bringing up the issue, @RajaVamsi11! We're gonna have to close this PR due to inactivity and merge conflicts but we're still working to resolve the issue.

PR #436, opened 2019-03-08
Move new user role ids into Constants class

### LINKED ISSUE #412: Move new user role ids into Constants class
**Summary:**

New users get role ids that are hardcoded. These should be moved to Constants class so it is made clear that they need to be changed together with other values in Constants.

aasaru:
Fixes #412

### DIFF
--- a/core/src/main/java/org/mifos/mobilewallet/core/domain/model/user/NewUser.java
+++ b/core/src/main/java/org/mifos/mobilewallet/core/domain/model/user/NewUser.java
@@ -3,6 +3,8 @@ package org.mifos.mobilewallet.core.domain.model.user;
 import java.util.ArrayList;
 import java.util.List;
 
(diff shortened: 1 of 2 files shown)


RajaVamsi11:
@aasaru Why did you closed #413 and #435. Try to solve the issue within one PR only.

aasaru:
Sorry about creating new pull requests.

RajaVamsi11:
@aasaru I am just asking to help you. No need to say sorry. If you face any issue regarding PR you can ask in the gitter channel.

aasaru:
@RajaVamsi11 ok, I didn't know there was a gitter channel, this would be most helpful, I will join that. I created new pull requests as I didn't know how to trigger travis to rebuild when I switched the target branch from master to dev. But then I realized that the build was failing for other reasons. Thanks for pointing out, next time I will stick with single PR.

PR #527, opened 2019-03-27
fix : Redundant modifiers on interfaces or interface components

### LINKED ISSUE #526: Unnecessary Interface modifiers
**Summary:**

While going through the codebase, I found there are many redundant modifiers on interfaces or interface components like modifier  public to the methods inside interface and static to the interface.

These kind of modifies can be removed as they are of no use. My idea is to remove the unnecessary interface modifiers from the project as the fields in an interface are implicitly public static final and the methods in an interface are by default public.

RajaVamsi11:
Fixes #526 


shiv07tiwari:
@RajaVamsi11 shifting to kotlin will fix this warning.

PR #671, opened 2020-01-25
fixes issue #670 , app crashes on pausing edit profile activity in dev branch.

### LINKED ISSUE #670: fix : App crashes when pausing EditProfile Activity in dev branch.
**Summary:** In Edit Profile Activity (of dev branch), when user clicks on change password and change passcode button or simply press back or home button, app crashes due to dialogBox.dismiss() in onPuase() method.

This issue is in dev branch as it is using our class dialogbox. Master branch works fine since it is using inner class.
When we create DialogBox dialogBox = new DialogBox(); object in edit profile activity. It creates the object of Alert Dialog in DialogBox class.
and when we click on change password/passcode button, we are checking whether dialogBox is null or not. It is not null, but since we didn't created a dialog hence the object alertDialog in DialogBox class is null. Therefore it is crashing the app on calling dialogBox.dismiss(). In background it is calling dismiss() method on null alertDialog object.

**Steps to reproduce:** Install dev branch app and click on change password or change passcode button, or back or home button while in Edit Profile Activity.

**Device and Android version:** Every device.

jatinsachdevame:
Fixes #670 

### DIFF
--- a/core/src/main/java/org/mifos/mobilewallet/core/data/fineract/api/BaseURL.java
+++ b/core/src/main/java/org/mifos/mobilewallet/core/data/fineract/api/BaseURL.java
@@ -8,11 +8,11 @@ public class BaseURL {
 
     public static final String PROTOCOL_HTTPS = "https://";
 
-    public static final String API_ENDPOINT = "demo.openmf.org";
+    public static final String API_ENDPOINT = "demo.mifos.io";
     public static final String API_PATH = "/fineract-provider/api/v1/";
 
     //self service url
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/mifos/mobilewallet/core/data/fineract/api/BaseURL.java
```diff
@@ -8,11 +8,11 @@
```
991rajat: Don't add this file in commit. This may help in not failing the build.

jatinsachdevame:
@991rajat Thanks buddy, I was wondering that why Build failed.

PR #681, opened 2020-01-26
fix: confirmation dialogBox for Delete Bank

### LINKED ISSUE #456: Add a confirmation Dialog in Delete Bank
**Summary:**

Create a confirmation dialog in delete Bank

laxyapahuja:
## Issue Fix
Fixes #456 

## Description
I've added a confirmation Dialog Box for Delete Bank using the ```DialogBox``` util class.

##

### DIFF
--- a/mifospay/src/main/java/org/mifos/mobilewallet/mifospay/bank/ui/BankAccountDetailActivity.java
+++ b/mifospay/src/main/java/org/mifos/mobilewallet/mifospay/bank/ui/BankAccountDetailActivity.java
@@ -1,6 +1,7 @@
 package org.mifos.mobilewallet.mifospay.bank.ui;
 
 import android.app.Activity;
+import android.content.DialogInterface;
 import android.content.Intent;
 import android.os.Bundle;
 import android.support.v7.widget.CardView;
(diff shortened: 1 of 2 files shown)


rachittt:
the dialog message "this will delete the selected bank." seems unnecessary in this case.

laxyapahuja:
> the dialog message "this will delete the selected bank." seems unnecessary in this case.

The DialogBox class has a message as a necessity. I can update it to accept null values. Thoughts? @luckyman20

luckyman20:
@laxyapahuja I think there can be one more constructor which doesn't requires `message`.

rachittt:
@laxyapahuja can you please update the PR and remove the dialogBox message.

rchtgpt (changes requested):
@laxyapahuja please remove the dialogBox message.

s-ayush2903:
Just a suggestion: It would be better if we have something like generic dialog box and use it wherever needed rather creating a fresh one all the time

rchtgpt:
@s-ayush2903, we have it already. 

@laxyapahuja please use it. [this](https://github.com/openMF/mobile-wallet/blob/dev/mifospay/src/main/java/org/mifos/mobilewallet/mifospay/utils/DialogBox.java) might help you.

rchtgpt:
@laxyapahuja if we don't get a response within 2 days, we will have to close this PR because it is restricting other contributors from working on it.

PR #717, opened 2020-02-08
 fix: app crash when user ddint enter a mobile number

### LINKED ISSUE #703: Fix: App crash when mobile number is not enetered in register activity
**Summary:**

When the user does not enter any mobile number under customer registration, then the app crashes instead of going back to the previous activity which should not

Bhavnaharitsa:
Fixes #703 
Fixed app crash when mobile number not entered in register activity

### DIFF
--- a/mifospay/src/main/java/org/mifos/mobilewallet/mifospay/registration/ui/MobileVerificationActivity.java
+++ b/mifospay/src/main/java/org/mifos/mobilewallet/mifospay/registration/ui/MobileVerificationActivity.java
@@ -73,7 +73,6 @@ public class MobileVerificationActivity extends BaseActivity implements
 
     @OnClick(R.id.btn_get_otp)
     public void onGetOTp() {
-        Utils.hideSoftKeyboard(this);
         if (mCcpCode.isValidFullNumber()) {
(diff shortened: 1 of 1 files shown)


naman14:
Can you explain how does this fix the crash? And what what the cause of the crash?

Also try to add the stacktrace in the corresponding issue also. It helps anyone reviewing having a context of the exact issue.

Bhavnaharitsa:
The User.hidesoftkeyboard was apparently causing the error. And the logs were showing quite the same. I had to remove it and the other line was showing error before the app run. I see, there is not much purpose of it and removing them solves(d) the problem. @naman14 
Logs, I shall share soon. Have a bit of problem with my studio.

naman14:
@Bhavnaharitsa were you able to include the stacktrace? If we don't know the root cause of the issue then I can't say that this will fix the issue

Bhavnaharitsa:
I will add it soon

Bhavnaharitsa:
@naman14 Here are the logs
image

devansh-299:
@Bhavnaharitsa after your changes even after entering OTP, the Floating action bar doesn't appear. Please fix this

luckyman20:
@Bhavnaharitsa Closing this because of inactivity.

PR #721, opened 2020-02-16
 fix:added function to clear text in search bar

### LINKED ISSUE #486: Clear the text of search menu.
**Summary:**

There should be a cross(x) button present in the search menu which clears the text in the search bar. It is currently missing.

Bhavnaharitsa:
Fixes #486 
Added function to clear the text when clicked on the cross mark. Had already made a pr but had to close it for some reason.

### DIFF
--- a/mifospay/src/main/java/org/mifos/mobilewallet/mifospay/bank/ui/LinkBankAccountActivity.java
+++ b/mifospay/src/main/java/org/mifos/mobilewallet/mifospay/bank/ui/LinkBankAccountActivity.java
@@ -37,6 +37,7 @@ import javax.inject.Inject;
 
 import butterknife.BindView;
 import butterknife.ButterKnife;
+import butterknife.OnClick;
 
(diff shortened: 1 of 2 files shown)


mifospay/src/main/java/org/mifos/mobilewallet/mifospay/bank/ui/LinkBankAccountActivity.java
```diff
@@ -57,6 +58,10 @@
     PopularBankAdapter mPopularBankAdapter;
     @Inject
     OtherBankAdapter mOtherBankAdapter;
+    @OnClick(R.id.et_cross_bank)
```
naman14: functions are not supposed to be here, please move this inside the class

Bhavnaharitsa: Sure

PR #727, opened 2020-02-17
Fix #726 - Added tests so that only fields of the section which is selected are checked

### LINKED ISSUE #726: Empty field validation in SendFragment does not work properly
**Summary:**

The logic for checking empty fields does not work work it allows submission of empty vpa field when mobile number field is filled and vice versa. 

**Steps to reproduce:**

Go to `SendFragment` then go the mobile segment fill all the fields return to the vpa section and try to submit with vpa field as empty.

**Screen Recording:**

dubeyaayush07:
Fixes #726


naman14:
Looks good

PR #737, opened 2020-02-21
Fix #729 - Added option to filter payment history

### LINKED ISSUE #729: Feat: option to filter transactions in payment history
**Summary:**
We can filter history based on whether the transactions are credit or debit we can also filter such that only transactions whose value is above a certain threshold are displayed

dubeyaayush07:
Fixes #729

### DIFF
--- a/core/src/main/java/org/mifos/mobilewallet/core/data/fineract/entity/mapper/TransactionMapper.java
+++ b/core/src/main/java/org/mifos/mobilewallet/core/data/fineract/entity/mapper/TransactionMapper.java
@@ -2,12 +2,15 @@ package org.mifos.mobilewallet.core.data.fineract.entity.mapper;
 
 import org.mifos.mobilewallet.core.data.fineract.entity.accounts.savings.SavingsWithAssociations;
 import org.mifos.mobilewallet.core.data.fineract.entity.accounts.savings.Transactions;
+import org.mifos.mobilewallet.core.domain.model.CheckBoxStatus;
 import org.mifos.mobilewallet.core.domain.model.Transaction;
 import org.mifos.mobilewallet.core.domain.model.TransactionType;
 import org.mifos.mobilewallet.core.utils.DateHelper;
 
 import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 
 import javax.inject.Inject;
 
@@ -25,17 +28,34 @@ public class TransactionMapper {
     }
 
     public List<Transaction> transformTransactionList(SavingsWithAssociations
-            savingsWithAssociations) {
+            savingsWithAssociations, List<CheckBoxStatus> filterList) {
+        List<Transaction> transactionList = new ArrayList<>();
+        Set<TransactionType> checkedFilterSet = getCheckedFilterSet(filterList);
+
+        if (savingsWithAssociations != null && savingsWithAssociations.getTransactions() != null
+                && savingsWithAssociations.getTransactions().size() != 0) {
+
+            for (Transactions transactions : savingsWithAssociations.getTransactions()) {
+                Transaction transaction = transformInvoice(transactions);
+                if (checkedFilterSet.contains(transaction.getTransactionType())) {
+                    transactionList.add(transaction);
+                }
+            }
+        }
+        return transactionList;
+    }
+
+    public List<Transaction> transformTransactionList(SavingsWithAssociations
+                                                              savingsWithAssociations) {
         List<Transaction> transactionList = new ArrayList<>();
 
         if (savingsWithAssociations != null && savingsWithAssociations.getTransactions() != null
                 && savingsWithAssociations.getTransactions().size() != 0) {
 
             for (Transactions transactions : savingsWithAssociations.getTransactions()) {
-                transactionList.add(transformInvoice(transactions));
+                Transaction transaction = transformInvoice(transactions);
+                transactionList.add(transaction);
             }
-
-
         }
         return transactionList;
     }
@@ -73,5 +93,21 @@ public class TransactionMapper {
             }
         }
         return transaction;
+    }
+
+    public Set<TransactionType> getCheckedFilterSet(List<CheckBoxStatus> filterList) {
+        Set<TransactionType> checkedFilterSet = new HashSet<>();
+        for (CheckBoxStatus status : filterList) {
+            if (status.getChecked()) {
+                if (status.getText().equals("Credit")) {
+                    checkedFilterSet.add(TransactionType.CREDIT);
+                } else if (status.getText().equals("Debit")) {
+                    checkedFilterSet.add(TransactionType.DEBIT);
+                } else if (status.getText().equals("Other")) {
+                    checkedFilterSet.add(TransactionType.OTHER);
+                }
+            }
+        }
+        return  checkedFilterSet;
     }
 }
--- a/core/src/main/java/org/mifos/mobilewallet/core/domain/model/CheckBoxStatus.java
+++ b/core/src/main/java/org/mifos/mobilewallet/core/domain/model/CheckBoxStatus.java
@@ -1,0 +1,27 @@
+package org.mifos.mobilewallet.core.domain.model;
+
+public class CheckBoxStatus {
+
+    private String text;
+    private boolean isChecked = false;
+
+    public CheckBoxStatus(String text) {
+        this.text = text;
+    }
+
+    public String getText() {
+        return text;
+    }
+
+    public void setText(String text) {
+        this.text = text;
+    }
+
+    public Boolean getChecked() {
+        return isChecked;
+    }
+
+    public void setChecked(Boolean checked) {
+        isChecked = checked;
+    }
+}
--- a/mifospay/src/main/java/org/mifos/mobilewallet/mifospay/history/HistoryContract.java
+++ b/mifospay/src/main/java/org/mifos/mobilewallet/mifospay/history/HistoryContract.java
@@ -1,6 +1,7 @@
 package org.mifos.mobilewallet.mifospay.history;
 
 import org.mifos.mobilewallet.core.data.fineract.entity.accounts.savings.TransferDetail;
+import org.mifos.mobilewallet.core.domain.model.CheckBoxStatus;
 import org.mifos.mobilewallet.core.domain.model.Transaction;
 import org.mifos.mobilewallet.mifospay.base.BasePresenter;
 import org.mifos.mobilewallet.mifospay.base.BaseView;
@@ -35,7 +36,7 @@ public interface HistoryContract {
 
     interface TransactionsHistoryPresenter extends BasePresenter {
 
-        void fetchTransactions();
+        void fetchTransactions(List<CheckBoxStatus> filterList);
 
         void handleTransactionClick(int transactionIndex);
 
--- a/mifospay/src/main/java/org/mifos/mobilewallet/mifospay/history/presenter/HistoryPresenter.java
+++ b/mifospay/src/main/java/org/mifos/mobilewallet/mifospay/history/presenter/HistoryPresenter.java
@@ -5,6 +5,7 @@ import org.mifos.mobilewallet.core.base.UseCase;
 import org.mifos.mobilewallet.core.base.UseCaseFactory;
 import org.mifos.mobilewallet.core.base.UseCaseHandler;
 import org.mifos.mobilewallet.core.domain.model.Account;
+import org.mifos.mobilewallet.core.domain.model.CheckBoxStatus;
 import org.mifos.mobilewallet.core.domain.model.Transaction;
 import org.mifos.mobilewallet.core.domain.usecase.account.FetchAccount;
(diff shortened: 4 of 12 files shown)


dubeyaayush07:
@naman14 could you review this?

core/src/main/java/org/mifos/mobilewallet/core/domain/model/CheckBoxStatus.java
```diff
@@ -0,0 +1,27 @@
+package org.mifos.mobilewallet.core.domain.model;
```
naman14, line 1: UI models should not be a part of core module. Core module should only have the usecase related entities

dubeyaayush07: Okay I will correct it

core/src/main/java/org/mifos/mobilewallet/core/domain/model/CheckBoxStatus.java
```diff
@@ -0,0 +1,27 @@
...
+
+public class CheckBoxStatus {
+
+    private String status;
```
naman14: not sure what status signifies here. variable name shouldn't be `status`. Can be changed to simply be `text`

mifospay/src/main/java/org/mifos/mobilewallet/mifospay/history/ui/HistoryFragment.java
```diff
@@ -62,10 +72,20 @@
     @BindView(R.id.pb_history)
     ProgressBar pbHistory;
 
+    private RecyclerView checkBoxRecyclerView;
+    private List<Transaction> transactions;
+    private List<CheckBoxStatus> statusList;
```
naman14: should be `filterList`

mifospay/src/main/java/org/mifos/mobilewallet/mifospay/history/ui/HistoryFragment.java
```diff
@@ -76,11 +96,83 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container,
         mPresenter.attachView(this);
 
         setupUi();
+        setStatusList();
```
naman14: `setFilterList`

mifospay/src/main/java/org/mifos/mobilewallet/mifospay/history/ui/HistoryFragment.java
```diff
@@ -62,10 +72,20 @@
     @BindView(R.id.pb_history)
     ProgressBar pbHistory;
 
+    private RecyclerView checkBoxRecyclerView;
```
naman14: can be made local, no need to store as a global variable

mifospay/src/main/java/org/mifos/mobilewallet/mifospay/history/ui/HistoryFragment.java
```diff
@@ -76,11 +96,83 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container,
...
+    public void onCreateOptionsMenu(Menu menu, MenuInflater inflater) {
+        super.onCreateOptionsMenu(menu, inflater);
+        inflater.inflate(R.menu.menu_filter, menu);
+    }
+
+    @Override
+    public boolean onOptionsItemSelected(MenuItem item) {
+        switch (item.getItemId()) {
+            case R.id.menu_filter_history:
+                showFilterDialog();
```
naman14, line 107: A good idea is to check for `getActivity() != null` as the options menu is inflated as a part of activity and is independent of the fragment's UI. Fragment might have been detached but the options menu might still be there. Clicking in suich cases will result in crash. Can also check if fragment is still attached by using `isAdded()`

mifospay/src/main/java/org/mifos/mobilewallet/mifospay/history/ui/HistoryFragment.java
```diff
@@ -62,10 +72,20 @@
     @BindView(R.id.pb_history)
     ProgressBar pbHistory;
 
+    private RecyclerView checkBoxRecyclerView;
+    private List<Transaction> transactions;
```
naman14: we don't need to store the transactions here

mifospay/src/main/java/org/mifos/mobilewallet/mifospay/history/HistoryContract.java
```diff
@@ -28,6 +28,8 @@
 
         void showTransactions(List<Transaction> transactions);
 
+        void setTransactions(List<Transaction> transactions);
```
naman14: no need for this, should be able to use the existing showTransactions

mifospay/src/main/java/org/mifos/mobilewallet/mifospay/history/presenter/HistoryPresenter.java
```diff
@@ -90,6 +95,37 @@ public void handleTransactionClick(int transactionIndex) {
         mHistoryView.showTransactionDetailDialog(transactionIndex, accountNumber);
     }
 
+    public List<Transaction> getFilteredTransactions(List<Transaction> transactions,
```
naman14: IMO the filtering can be made part of code usecase. Instead of adding a new `getFiilteredTransactions`, pass a filterList argument in existing `fetchTransactions` and then make filtering a part of the core usecase.

Let me know if I should provide some more clarity regarding this.

mifospay/src/main/java/org/mifos/mobilewallet/mifospay/history/ui/adapter/CheckBoxAdapter.java
```diff
@@ -0,0 +1,89 @@
...
+                R.layout.row_checkbox, parent, false);
+        return new ViewHolder(v);
+    }
+
+    @Override
+    public void onBindViewHolder(RecyclerView.ViewHolder holder, int position) {
+
+        CheckBoxStatus statusModel = statusList.get(position);
+
+        int[][] states = {{android.R.attr.state_checked}, {}};
```
naman14: not using this, can be removed

mifospay/src/main/java/org/mifos/mobilewallet/mifospay/history/ui/adapter/CheckBoxAdapter.java
```diff
@@ -0,0 +1,89 @@
...
+
+import javax.inject.Inject;
+
+import butterknife.BindView;
+import butterknife.ButterKnife;
+import butterknife.OnCheckedChanged;
+import butterknife.OnClick;
+
+public class CheckBoxAdapter extends RecyclerView.Adapter<RecyclerView.ViewHolder> {
+    private Context context;
```
naman14: we dont need context here

mifospay/src/main/java/org/mifos/mobilewallet/mifospay/history/presenter/HistoryPresenter.java
```diff
@@ -90,6 +95,37 @@ public void handleTransactionClick(int transactionIndex) {
...
+                                                     List<CheckBoxStatus> statusList) {
+        ArrayList<Transaction> filteredTransactions = new ArrayList<>();
+        List<CheckBoxStatus> CheckBoxStatusList = getCheckedStatusList(statusList);
+        for (Transaction transaction : transactions) {
```
naman14: avoid nested for loops especially more so because number of transactions can get very large.

mifospay/src/main/java/org/mifos/mobilewallet/mifospay/history/presenter/HistoryPresenter.java
```diff
@@ -90,6 +95,37 @@ public void handleTransactionClick(int transactionIndex) {
...
+            for (CheckBoxStatus status : CheckBoxStatusList) {
+                if (status.getStatus().compareTo(Constants.CREDIT) == 0 &&
```
naman14: can use `.equals` instead of compareTo (we don't have a custom comparator anyway in CheckBoxStatus)

mifospay/src/main/java/org/mifos/mobilewallet/mifospay/history/presenter/HistoryPresenter.java
```diff
@@ -90,6 +95,37 @@ public void handleTransactionClick(int transactionIndex) {
...
+                        transaction.getTransactionType() == TransactionType.OTHER) {
+                    filteredTransactions.add(transaction);
+                }
+            }
+        }
+        return filteredTransactions;
+    }
+
+    private List<CheckBoxStatus> getCheckedStatusList(List<CheckBoxStatus> statusList) {
+        ArrayList<CheckBoxStatus> CheckBoxStatusList = new ArrayList<>();
```
naman14: camelcase

naman14:
Hey @dubeyaayush07 thanks for this PR. Seems like a good functionality to add. I had 1 comment regarding the filter logic (it should be a part of usecase so that it's not bound to the UI). Why? This is where clean architecture helps us. This allows us to plug filters dynamically later on. Maybe we start sending filterlist in api instead of filtering locally after we have already fetched all transactions. Maybe we need to save the currently selected filters in which case when users comes next time, we should only be fetching the selected filters. We also make sure that other modules are also able to use filtering if its a part of core module.

Also, maybe we can also add it at the api level. You may need to check the api docs if it supports filtering there. (Can be a separate PR) but make sure to make this a part of core usecase.

Let me know if you need more clarity.

dubeyaayush07:
@naman14 So instead of filtering in the presenter I should modify fetchTransactions(maybe overload the method ) to use filterList. Am I understanding it correctly?

naman14:
Yes, pass filterList to fetchTransactions (can overload) and then pass to usecase where filtering will take place

mifospay/src/main/java/org/mifos/mobilewallet/mifospay/history/presenter/HistoryPresenter.java
dubeyaayush07: Thanks for the feedback, I agree the nested for loops look ugly but my thought process while doing this was that for a particular filtering process the number of checkboxes or filter parameters are constant so the runtime scales linearly in the number of transactions making it O(n).

dubeyaayush07:
@naman14 I modified the code such that the filtering now takes place in the Transaction Mapper, I was nervous about changing code other than that of `HistoryFragment` so I overloaded the `fetchTransactionsHistory` method. The credit, debit constants were not accessible from the the core module so for now I have hardcoded string literals.

core/src/main/java/org/mifos/mobilewallet/core/data/fineract/entity/mapper/TransactionMapper.java
```diff
@@ -25,17 +28,34 @@ public TransactionMapper() {
...
+                Transaction transaction = transformInvoice(transactions);
+                if (checkedFilterSet.contains(transaction.getTransactionType())) {
+                    transactionList.add(transaction);
+                }
             }
+        }
+        return transactionList;
+    }
 
+    public List<Transaction> transformTransactionList(SavingsWithAssociations
```
shiv07tiwari, line 48: Hey @dubeyaayush07 , here you have used method overloading but that also has increased the redundant code. I'll suggest a better way for this would be to pass a null filter list from this function whenever transformTransactionList is called with 1 argument, to the actual transformTransactionList, and modify the implementation accordingly.

core/src/main/java/org/mifos/mobilewallet/core/domain/model/CheckBoxStatus.java
shiv07tiwari, line 1: Please remove this from the core module

mifospay/src/main/java/org/mifos/mobilewallet/mifospay/history/ui/adapter/CheckBoxAdapter.java
```diff
@@ -0,0 +1,84 @@
...
+
+        View v = LayoutInflater.from(parent.getContext()).inflate(
+                R.layout.row_checkbox, parent, false);
+        return new ViewHolder(v);
+    }
+
+    @Override
+    public void onBindViewHolder(RecyclerView.ViewHolder holder, int position) {
+
+        CheckBoxStatus statusModel = statusList.get(position);
```
shiv07tiwari, line 41: Remove the extra unnecessary spaces.

shiv07tiwari (changes requested):

shiv07tiwari:
@dubeyaayush07 any updates on this?

dubeyaayush07:
@shiv07tiwari I am busy with my exams right now, I will try to make the necessary updates as soon as I get time

dubeyaayush07:
@devansh-299 @shiv07tiwari It seems like the filtering functionality has already been implemented by someone else can I close this PR?

dubeyaayush07:
@shiv07tiwari @devansh-299 @naman14  In this pull request I was using a checkbox menu for the filter list but since my last commit chips were added for selecting transaction type. Should I use the menu or the chips?

PR #771, opened 2020-02-27
Fix #770 - improved currency formatting

### LINKED ISSUE #770: Improve Currency Formatting
**Summary:**

Currently currency is formatted in two parts  first the amount is formatted with decimal formatter then it is concatenated with the currency code what I am proposing is that we should use number formatter with its currency set using the currency code of the transaction this way we will be able to use currency specific formatting and symbol.

dubeyaayush07:
Fix #770
Currencies are now formatted by passing currency code and amount to the utils function which formats these currencies according to their currency code and also gives them an appropriate currency symbol

Screenrecorder-2020-02-27-12-36

### DIFF
--- a/mifospay/src/main/java/org/mifos/mobilewallet/mifospay/utils/Utils.java
+++ b/mifospay/src/main/java/org/mifos/mobilewallet/mifospay/utils/Utils.java
@@ -7,7 +7,8 @@ import android.view.Display;
 import android.view.WindowManager;
 import android.view.inputmethod.InputMethodManager;
 
-import java.text.DecimalFormat;
+import java.text.NumberFormat;
+import java.util.Currency;
 
 /**
  * Created by naman on 17/6/17.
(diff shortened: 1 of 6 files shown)


devansh-299:
@shiv07tiwari @luckyman20 please have a look

mifospay/src/main/java/org/mifos/mobilewallet/mifospay/utils/Utils.java
```diff
@@ -41,8 +42,11 @@ public static boolean isBlank(String s) {
         return true;
     }
 
-    public static String getFormattedAccountBalance(Double balance) {
-        DecimalFormat accountBalanceFormatter = new DecimalFormat("#,###,###.##");
+    public static String getFormattedAccountBalance(Double balance, String currencyCode) {
+        NumberFormat accountBalanceFormatter = NumberFormat.getCurrencyInstance();
+        accountBalanceFormatter.setMaximumFractionDigits(0);
+        accountBalanceFormatter.setCurrency(Currency.getInstance(currencyCode));
+
```
luckyman20, line 49:
```suggestion
```

PR #775, opened 2020-02-28
Fix #774 - bottom navigation bar does not move up with the keyboard

### LINKED ISSUE #774: Bottom Navigation bar moves up with keyboard
**Summary:**

When inputting values by using keyboard the bottom navigation bar moves up cluttering the UI. I have observed this behaviour does not happen always for example when the server is not responding the navbar stays down but still it's behaviour is finicky.

dubeyaayush07:
Fix #774

## Screenshots

### DIFF
--- a/mifospay/src/main/AndroidManifest.xml
+++ b/mifospay/src/main/AndroidManifest.xml
@@ -29,6 +29,7 @@
         </activity>
         <activity
             android:name=".home.ui.MainActivity"
+            android:windowSoftInputMode="adjustPan"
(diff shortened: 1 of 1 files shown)


dubeyaayush07:
While searching the internet for a solution I found the only way to fix this problem was by adding `android:windowSoftInputMode="adjustPan"` to the manifest but adding the above line had some side effect for people who were using scrollview. Our app does not use scroll view in the main activity and I have verified that the app is working fine after adding this line but I just want the maintainers to be aware of this.

Issue #728, opened 2020-02-18
fix: Show QrActivity does not show any message when no value is enteres in reset amount


Bhavnaharitsa:
**Summary:**

The ShowQrActivity accepts even empty value, when the user resets the amount. It should not accept. It should show an error or some toast saying the user cannot enter empty amount

PR #781, opened 2020-02-29
fix: shows message when user doesnt enter any amount

Bhavnaharitsa:
Fixes: #728
Altered the condition when QRActivity shows pop up message when no amount is displayed/given


mifospay/src/main/java/org/mifos/mobilewallet/mifospay/qr/ui/ShowQrActivity.java
```diff
@@ -90,7 +90,7 @@ void showSetAmountDialog (final String qrData) {
             @Override
             public void onClick(DialogInterface dialog, int which) {
                 String amount = edittext.getText().toString();
-                if (amount.equals("")) {
+                if (amount.isEmpty()) {
```
dubeyaayush07: Empty amount validation is working fine and why change equals with isEmpty they both accomplish the same thing

Bhavnaharitsa: It is not which is why I opened the issue. Apparently, I don't know why, that works for me. Let me attach a screen shot

Issue #784, opened 2020-03-02
feat: Refurbish Settings page


Bhavnaharitsa:
**Summary:**

Our settings page is quite noobish and does not follow any conventional app. We need to make one like that.
I suggest to refurbish the entire settings page using PreferenceCategory in Android and make a new settings page with the same functionalities. 
Here is the link about it:
https://developer.android.com/reference/android/preference/PreferenceCategory

AntonAsmirko:
can I work on this issue?

haran2248:
Is this taken up?Can i work on this issue?
@devansh-299

therajanmaurya:
Fixed in compose implementation

PR #785, opened 2020-03-02
Feat: refurbished the entire setting page

Bhavnaharitsa:
Feat: #784
Refurbished the entire settings page, with the PreferenceCategory. Used the preference_category in xml and updated the fragment and called in activity

Working video:
ezgif com-video-to-gif

### DIFF
--- a/mifospay/src/main/res/values/strings.xml
+++ b/mifospay/src/main/res/values/strings.xml
@@ -132,11 +132,11 @@
     <string name="link_your_bank_account">Link Your Bank Account</string>
     <string name="no_linked_bank_accounts">No Linked Bank Accounts</string>
     <string name="editable_image_hint">Editable image hint</string>
-    <string name="change_password">Change Password</string>
+    <string name="change_password">Changes your Account Password</string>
     <string name="current_password">Current Password</string>
     <string name="new_password">New Password</string>
     <string name="confirm_new_password">Confirm New Password</string>
-    <string name="change_passcode">Change Passcode</string>
+    <string name="change_passcode">Changes your Account Passcode</string>
     <string name="current_passcode">Current Passcode</string>
     <string name="new_passcode">New Passcode</string>
     <string name="confirm_new_passcode">Confirm New Passcode</string>
@@ -248,5 +248,15 @@
     <string name="grip">Grip</string>
     <string name="hide_balance">Hide Balance</string>
     <string name="no_bank_account">No bank found</string>
+    <string name="language_type">Choose your language</string>
 
+    <!-- TODO: Remove or change this placeholder text -->
+    <string name="passcode">Passcode</string>
+    <string name="other">Other</string>
+    <string name="choose_language">Choose your language</string>
+    <string name="language">Language</string>
+    <string name="pref_configuration">Pref Configuration</string>
+    <string name="pref_base_url_title">Update Endpoint</string>
+    <string name="pref_base_url_desc">Click here to change your Endpoint Configurations</string>
+    <string name="disable_account_desc">Disables your Current Account</string>
 </resources>
--- a/mifospay/src/main/res/values/styles_frame.xml
+++ b/mifospay/src/main/res/values/styles_frame.xml
@@ -1,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+
+    <style name="FrameLayout"/>
+
+    <style name="FrameLayout.Base">
+        <item name="android:layout_width">match_parent</item>
+        <item name="android:layout_height">match_parent</item>
+    </style>
+
+    <style name="FrameLayout.Base.Container">
+        <item name="android:id">@id/container</item>
+    </style>
+
+</resources>
--- a/mifospay/src/main/res/values/styles_linear.xml
+++ b/mifospay/src/main/res/values/styles_linear.xml
@@ -1,0 +1,23 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+
+    <style name="LinearLayout"/>
+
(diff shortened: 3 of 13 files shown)


Bhavnaharitsa:
@naman14 @luckyman20 @shiv07tiwari @droidchef Please review this!!

mifospay/src/main/res/values/strings.xml
```diff
@@ -132,11 +132,11 @@
     <string name="link_your_bank_account">Link Your Bank Account</string>
     <string name="no_linked_bank_accounts">No Linked Bank Accounts</string>
     <string name="editable_image_hint">Editable image hint</string>
-    <string name="change_password">Change Password</string>
+    <string name="change_password">Changes your Account Password</string>
     <string name="current_password">Current Password</string>
     <string name="new_password">New Password</string>
     <string name="confirm_new_password">Confirm New Password</string>
-    <string name="change_passcode">Change Passcode</string>
+    <string name="change_passcode">Changes your Account Passcode</string>
```
jawidMuhammadi: I think it should be `change` not `changes`

jawidMuhammadi (changes requested):

mifospay/src/main/res/values/strings.xml
```diff
@@ -248,5 +248,16 @@
     <string name="grip">Grip</string>
     <string name="hide_balance">Hide Balance</string>
     <string name="no_bank_account">No bank found</string>
+    <string name="language_type">Choose your language</string>
 
+    <!-- TODO: Remove or change this placeholder text -->
+    <string name="hello_blank_fragment">Hello blank fragment</string>
```
jawidMuhammadi: Delete this string

jawidMuhammadi (changes requested):

mifospay/src/main/res/values/styles_frame.xml
```diff
@@ -0,0 +1,15 @@
...
+<resources>
+
+    <style name="FrameLayout"/>
+
+    <style name="FrameLayout.Base">
+        <item name="android:layout_width">match_parent</item>
+        <item name="android:layout_height">match_parent</item>
+    </style>
+
+    <style name="FrameLayout.Container" parent="FrameLayout.Base">
```
jawidMuhammadi: Instead of '<style name="FrameLayout.Container" parent="FrameLayout.Base">' it would be much better if you do 
'<style name="FrameLayout.Base.Container" >'

jawidMuhammadi (changes requested):

mifospay/src/main/res/values/styles_linear.xml
```diff
@@ -0,0 +1,23 @@
...
+        <item name="android:layout_height">match_parent</item>
+        <item name="android:orientation">vertical</item>
+    </style>
+
+    <style name="LinearLayout.Wrap">
+        <item name="android:layout_width">wrap_content</item>
+        <item name="android:layout_height">wrap_content</item>
+    </style>
+
+    <style name="LinearLayout.Width">
```
jawidMuhammadi: I think it is much better to define layout width and height in the layout itself, not in the style. Even by that are you sure `name="LinearLayout.Width"` is a general and a good name for your defined style which everyone can understand?

mifospay/src/main/res/values/strings.xml
Bhavnaharitsa: Okay, will change it

Bhavnaharitsa: Okay, will change it

mifospay/src/main/res/values/strings.xml
Bhavnaharitsa: It got when I created the fragment by default

mifospay/src/main/res/values/styles_linear.xml
Bhavnaharitsa: I think, it would be a global style which many would use in the future whenever they need the linear layout

mifospay/src/main/res/values/strings.xml
jawidMuhammadi: If there is no need for this string then please remove it.

Bhavnaharitsa: I will remove it :)

dubeyaayush07:
You forgot the logout button

Bhavnaharitsa:
@dubeyaayush07 Yess, thank you for noticing!

shiv07tiwari:
@Bhavnaharitsa Please add the updated screenshot.

Bhavnaharitsa:
@shiv07tiwari Sir, about that. Apparently the PreferenceFragment that I used is dedicated for android and only available in Androidx. So, can we, with your permission migrate to Androidx as we might loose a lot of libraries if we still sustain in Android.

shiv07tiwari:
@Bhavnaharitsa Any updates on this?

Bhavnaharitsa:
@shiv07tiwari @luckyman20 @rachittt As we all know, the settings page mobile wallet has is not like the conventional one in other apps which is the reason I opened this pr. But, what I have learned is that we don't use AndroidX and the Preference feature works on AndroidX alone as it has been deprecated in Android now. If there is any task as such to migrate the project to AndroidX, I will be happy to take up and convert this settings page too along with it :)

luckyman20:
@Bhavnaharitsa Adding androidX gives error with Dagger, last time I checked @shiv07tiwari was working on it.

haran2248:
@Bhavnaharitsa @luckyman20 @shiv07tiwari @devansh-299 Can i take up this issue?Is the pull request faulty?Please let me know

therajanmaurya:
Unrelated and not needed anymore

Issue #794, opened 2020-03-07
App crashes when null amount submitted in MerchantTransferActivity


singaltanmay:
**Summary:**

Due to no null check taking place before converting a String to a Double, the app crashes when the submit button is clicked.

**Steps to reproduce:**

In MerchantTransferActivity press the Submit button without entering any amount.

**Expected behavior:**

Prompt the user to enter an amount.

**Observed behavior:**

The app crashes. Logcat : 
```

2020-03-07 19:13:13.964 3325-3325/org.mifos.mobilewallet.mifospay E/AndroidRuntime: FATAL EXCEPTION: main
    Process: org.mifos.mobilewallet.mifospay, PID: 3325
    java.lang.NumberFormatException: empty String
        at sun.misc.FloatingDecimal.readJavaFormatString(FloatingDecimal.java:1842)
        at sun.misc.FloatingDecimal.parseDouble(FloatingDecimal.java:110)
        at java.lang.Double.parseDouble(Double.java:538)
        at org.mifos.mobilewallet.mifospay.merchants.ui.MerchantTransferActivity.makeTransaction(MerchantTransferActivity.java:146)
        at org.mifos.mobilewallet.mifospay.merchants.ui.MerchantTransferActivity_ViewBinding$1.doClick(MerchantTransferActivity_ViewBinding.java:45)
        at butterknife.internal.DebouncingOnClickListener.onClick(DebouncingOnClickListener.java:22)
        at android.view.View.performClick(View.java:7125)
        at android.view.View.performClickInternal(View.java:7102)
        at android.view.View.access$3500(View.java:801)
        at android.view.View$PerformClick.run(View.java:27336)
        at android.os.Handler.handleCallback(Handler.java:883)
        at android.os.Handler.dispatchMessage(Handler.java:100)
        at android.os.Looper.loop(Looper.java:214)
        at android.app.ActivityThread.main(ActivityThread.java:7356)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:492)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:930)

```
**Device and Android version:**

Android Studio Emulator API 29

dubeyaayush07:
duplicate #757 please check before submitting a new issue

Issue #757, opened 2020-02-24
No empty field validation in MerchantTransferActivity


dubeyaayush07:
**Summary:**
MerchantTransferActivity does not validate for empty amount field

dubeyaayush07:
I would like to work on this

luckyman20:
@dubeyaayush07 Please mention steps to reproduce.

dubeyaayush07:
@luckyman20 
Steps to Reproduce
Go to Merchant Transfer Activity do not enter anything in the amount field click submit the app crashes because null value is allowed

luckyman20:
How to reach to MerchantTransferActivity? Detailed description is needed.

dubeyaayush07:
![Screenrecorder-2020-03-17-16-37](https://user-images.githubusercontent.com/30518564/76850779-14084900-686e-11ea-8871-0d09ddce0f91.gif)
@luckyman20

luckyman20:
@dubeyaayush07 I am not able to load the merchants only. On which branch and server you are testing this?

dubeyaayush07:
@luckyman20  I am testing on the dev branch I removed KYC fragment from view layout to test this, view layout loads two fragments at once and when fetch fails in the KYC fragment it redirects to the HomeFragment that is why you are not able to access the merchant fragment. I am currently working on fixing this problem in another issue.

PR #758, opened 2020-02-24
Fix #757 - added empty field validation in MerchantTransferActivity

dubeyaayush07:
Fixes #757

- [x] Apply the `AndroidStyle.xml` style template to your code in Android Studio.

- [x] Run the unit tests with `./gradlew check` to make sure you didn't break anything

- [x] If you have multiple commits please combine them into one commit by squashing them.

PR #796, opened 2020-03-07
Fixed app crash when submitting null amount in MerchantTransferActivity

singaltanmay:
## Issue Fix
Fixes issues #794 and #757

## Description
In `org.mifos.mobilewallet.mifospay.merchants.ui.MerchantTransferActivity` when `R.id.btn_submit`  is clicked, `makeTransaction()` now checks if the amount fetched from the EditText is not null before parsing the String to a Double. It also shows the user a Toast if an amount has not been entered

##


dubeyaayush07:
There is already a PR for this issue (#758)

singaltanmay:
The earlier PR for this issue (#758) check only if the value is "" but not if it is null. [TextUtils.isEmpty()](https://developer.android.com/reference/android/text/TextUtils#isEmpty(java.lang.CharSequence)) is a superior check.  That PR also renames the variable "eamount"  to "amount".

dubeyaayush07:
@singaltanmay The getText function outputs an empty string when there is a null value, eamount is not grammatically correct also using equals to check for empty field is consistent with the codebase still if you want to suggest something do it in the original PR and the author will make the correction no need to create redundant PR as this creates extra work for the maintainers

Issue #816, opened 2020-03-14
fix: restrict amount in ShowQr


Bhavnaharitsa:
**Summary:**
The ShowQr, in the reset amount, the tag accepts as much as amount possible without any constraint. Usually, banks have a constraint of about one lakh transaction. I think we should have that too

PR #820, opened 2020-03-15
 fix: amount restricted when exceeds level

Bhavnaharitsa:
Fixes: #816
Amount restricted when exceeds level in ShowQrActivity.

### DIFF
--- a/mifospay/src/main/java/org/mifos/mobilewallet/mifospay/utils/Constants.java
+++ b/mifospay/src/main/java/org/mifos/mobilewallet/mifospay/utils/Constants.java
@@ -169,6 +169,7 @@ public class Constants {
     public static final String CHANGE_UPI_PIN = "Change UPI PIN";
     public static final String FORGOT_UPI_PIN = "Forgot UPI PIN";
 
+
     // broadcast receiver intent filters
     public static final String REGISTRATION_COMPLETE = "registrationComplete";
     public static final String PUSH_NOTIFICATION = "pushNotification";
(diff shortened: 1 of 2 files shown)


Bhavnaharitsa:
ezgif com-video-to-gif

mifospay/src/main/java/org/mifos/mobilewallet/mifospay/utils/Constants.java
```diff
@@ -168,6 +168,7 @@
     public static final String INDEX = "Index";
     public static final String CHANGE_UPI_PIN = "Change UPI PIN";
     public static final String FORGOT_UPI_PIN = "Forgot UPI PIN";
+    public static final String AMOUNT_EXCEED_LIMIT = "Amount exceeding limit";
```
jawidMuhammadi: Please do not use strings as constant if it is being displayed in UI, whenever we try to translate the app to other languages, if the strings are not in the `string.xml` file then we can't translate those strings and it will cause to very bad user experience.

Bhavnaharitsa: Okay sir!

Bhavnaharitsa: But, there are many strings that are in Constants, as you can see the top string is constant Do you want me to change that?

jawidMuhammadi: Yes please, in this PR at least change your added string,  the other developers mistakenly have added some UI related strings without thinking about future translation problems in Constants, so it doesn't mean we have to the same mistake. It would be great if you raise a refactor code issue and send another PR regarding transferring those strings to the `strings.xml` file that need the translation.

Bhavnaharitsa: Sure will, thank you for the guidance. I shall do the changes for now and open a new issue. However I have noticed a lot of hardcoded strings in Java files which needs to be changed too

Issue #819, opened 2020-03-15
fix: Many hardcoded strings in java files


Bhavnaharitsa:
**Summary:**
There are many hardcoded strings in many java files which can be assigned string values. 
For example:
1. ShowQr Activity
2. SignUpActivity
3. LinkBankAccountActivity
etc.

PR #821, opened 2020-03-16
fix: removed hard coded strings in java files and replaced them with strings

Bhavnaharitsa:
Fixes: #819
Removed all the hardcoded strings and converted them intro strings.xml strings

### DIFF
--- a/mifospay/src/main/java/org/mifos/mobilewallet/mifospay/bank/fragment/UpiPinFragment.java
+++ b/mifospay/src/main/java/org/mifos/mobilewallet/mifospay/bank/fragment/UpiPinFragment.java
@@ -74,7 +74,7 @@ public class UpiPinFragment extends BaseFragment implements BankContract.UpiPinV
         if (b != null) {
             step = b.getInt(Constants.STEP, 0);
             upiPin = b.getString(Constants.UPI_PIN, null);
-            mTvTitle.setText("Re-enter UPI PIN");
+            mTvTitle.setText(R.string.re_enter_pin);
         }
 
         mPeUpiPin.setOnEditorActionListener(new TextView.OnEditorActionListener() {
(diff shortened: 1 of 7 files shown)


mifospay/src/main/java/org/mifos/mobilewallet/mifospay/bank/fragment/UpiPinFragment.java
```diff
@@ -74,7 +74,7 @@ public View onCreateView(@NonNull LayoutInflater inflater, @Nullable ViewGroup c
         if (b != null) {
             step = b.getInt(Constants.STEP, 0);
             upiPin = b.getString(Constants.UPI_PIN, null);
-            mTvTitle.setText("Re-enter UPI PIN");
+            mTvTitle.setText(R.string.re_enter_pin);
```
jawidMuhammadi: Are you sure it is working as expected and displaying the correct text? As I see you are setting the id of the string as the text, I think inside the java code we have to do something like this `getString(R.string.re_enter_pin);`

mifospay/src/main/java/org/mifos/mobilewallet/mifospay/bank/fragment/UpiPinFragment.java
```diff
@@ -98,19 +98,19 @@ public void okayClicked() {
                     if (upiPin.equals(mPeUpiPin.getText().toString())) {
                         ((SetupUpiPinActivity) getActivity()).upiPinConfirmed(upiPin);
                     } else {
-                        showToast("UPI PIN mismatch");
+                        showToast(R.string.upi_pin_mismatch);
```
jawidMuhammadi: Same here and in all your other string changes.

shiv07tiwari:
@Bhavnaharitsa I agree with @jawidMuhammadi . Please fix it and also look into the Travis Build issue.

Bhavnaharitsa:
@shiv07tiwari Yes, will do the neeful

PR #838, opened 2020-03-19
fix: hidden add cards when cards aren't fetched

### LINKED ISSUE #834: Fix: Hide add cards when cards are not fetched
**Summary:**

When the existing cards are not fetched, I think we can hide the add card option because the user might not be sure what card he/she already has and might want to see the existing cards before adding a card.

Bhavnaharitsa:
Fix: #834 
Hidden cards when the cards aren't fetched

### DIFF
--- a/mifospay/src/main/java/org/mifos/mobilewallet/mifospay/savedcards/ui/CardsFragment.java
+++ b/mifospay/src/main/java/org/mifos/mobilewallet/mifospay/savedcards/ui/CardsFragment.java
@@ -3,6 +3,7 @@ package org.mifos.mobilewallet.mifospay.savedcards.ui;
 import android.content.res.Resources;
 import android.os.Bundle;
 import android.support.annotation.Nullable;
+import android.support.design.chip.Chip;
 import android.support.v4.widget.SwipeRefreshLayout;
 import android.support.v7.widget.DividerItemDecoration;
 import android.support.v7.widget.LinearLayoutManager;
@@ -67,6 +68,9 @@ public class CardsFragment extends BaseFragment implements CardsContract.CardsVi
     @Inject
     CardsAdapter mCardsAdapter;
 
+    @BindView(R.id.btn_add_card)
+    Chip addCard;
+
     View rootView;
 
     public static CardsFragment newInstance() {
@@ -103,6 +107,7 @@ public class CardsFragment extends BaseFragment implements CardsContract.CardsVi
             public void onRefresh() {
                 getSwipeRefreshLayout().setRefreshing(false);
                 mCardsPresenter.fetchSavedCards();
+                addCard.setVisibility(View.VISIBLE);
             }
         });
     }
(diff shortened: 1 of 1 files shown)


Bhavnaharitsa:
Screenshot_20200319-140626_Mifos Pay

devansh-299:
@Bhavnaharitsa I don't think there's a need to remove this. The user should be allowed to add cards. If any problem occurs in the adding process then it can be handled separately

mifospay/src/main/java/org/mifos/mobilewallet/mifospay/savedcards/ui/CardsFragment.java
```diff
@@ -125,6 +129,7 @@ private void showEmptyStateView() {
     public void showErrorStateView(int drawable, int title, int subtitle) {
         rvCards.setVisibility(View.GONE);
         pbCards.setVisibility(View.GONE);
+        addCard.setVisibility(View.GONE);
```
rahul-jha98: Since the data can be refreshed even when the user is on the current screen (due to swipe to refresh layout) and the cards may be populated then, make sure that the addCard chip is explicitly set to visible whenever the cards are populated.
This will prevent any inconsistent state in the UI.

rahul-jha98 (changes requested):

mifospay/src/main/java/org/mifos/mobilewallet/mifospay/savedcards/ui/CardsFragment.java
Bhavnaharitsa: Sure, I'll check that

Bhavnaharitsa:
@luckyman20 Please review

mifospay/src/main/java/org/mifos/mobilewallet/mifospay/savedcards/ui/CardsFragment.java
```diff
@@ -103,6 +107,7 @@ private void setUpSwipeRefresh() {
             public void onRefresh() {
                 getSwipeRefreshLayout().setRefreshing(false);
                 mCardsPresenter.fetchSavedCards();
+                addCard.setVisibility(View.VISIBLE);
```
devansh-299, line 110: Instead of making the `addCard` visible here, it would be much better to make it visible only when the `mCardsPresenter.fetchSavedCards()` the response is successful i.e. inside `showSavedCards` function

Bhavnaharitsa: Okay

Bhavnaharitsa:
@devansh-299 But, I'm unwell and down with fever. So I might take some time until I show results on this pr.

devansh-299:
> @devansh-299 But, I'm unwell and down with fever. So I might take some time until I show results on this pr.

@Bhavnaharitsa  No problem, take your time and get well!

devansh-299:
> @devansh-299 But, I'm unwell and down with fever. So I might take some time until I show results on this pr.

@Bhavnaharitsa can you please update this PR? if you not comfortable, someone else can work on the corresponding issue

Bhavnaharitsa:
@devansh-299 I'll do it asap

PR #859, opened 2020-03-25
Fixes #858 - Keyboard moves down when toggling from SendFragment to RequestFragment

### LINKED ISSUE #858: Keyboard stays up when toggling from SendFragment to RequestFragment
**Summary:**

When moving from SendFragment to RequestFramgment if the keyboard is up it does not go down.
Screenrecorder-2020-03-25-21-58

dubeyaayush07:
Fixes #858

Screenrecorder-2020-03-25-22-15

### DIFF
--- a/mifospay/src/main/java/org/mifos/mobilewallet/mifospay/utils/Utils.java
+++ b/mifospay/src/main/java/org/mifos/mobilewallet/mifospay/utils/Utils.java
@@ -16,11 +16,13 @@ import java.text.DecimalFormat;
 public class Utils {
 
     public static void hideSoftKeyboard(Activity activity) {
-        InputMethodManager inputMethodManager =
-                (InputMethodManager) activity.getSystemService(
-                        Activity.INPUT_METHOD_SERVICE);
-        inputMethodManager.hideSoftInputFromWindow(
-                activity.getCurrentFocus().getWindowToken(), 0);
+        if (activity.getCurrentFocus() != null) {
+            InputMethodManager inputMethodManager =
+                    (InputMethodManager) activity.getSystemService(
+                            Activity.INPUT_METHOD_SERVICE);
+            inputMethodManager.hideSoftInputFromWindow(
+                    activity.getCurrentFocus().getWindowToken(), 0);
+        }
     }
 
     public static int dp2px(Context context, int dp) {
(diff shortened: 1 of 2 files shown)


dubeyaayush07:
Hiding the keyboard in RequestFragment did not work as View pager loads the adjacent fragment in advance that is why I had to resort to hiding the keyboard via `addOnPageChangeListener` in Payment Fragment

dubeyaayush07:
@luckyman20 please review

dubeyaayush07:
> @dubeyaayush07 This throws a `NullPointerException` if the keyboard is not visible in the view (e.g. switching from `Request` tab to `History` tab). `addOnPageChangeListener` is called on every change of fragment, so you need to check first if the keyboard is already up in the first place, then you should deflate it.

Thanks I will update the PR

dubeyaayush07:
@sashank27 I found the cause of the `NullPointerException` was the use of `getCurrentFocus` in the Utils function to get the window token. When the focus was lost it produced a null value so to correct it I used the `PaymentsFragment'`s rootview to get the window token instead.  `hideSoftInputFromWindow` closes the keyborad in all situations.

sashank27:
To avoid the code reuse, I would suggest modifying `Utils.hideSoftKeyboard`, such that it makes sure that the `getCurrentFocus` does not produce a Null value, and keep the previous implementation in `PaymentsFragment`.
This would also handle a bug in login, related to keyboard inflation.

dubeyaayush07:
@luckyman20  @sashank27 please review

mifospay/src/main/java/org/mifos/mobilewallet/mifospay/utils/Utils.java
```diff
@@ -16,11 +16,13 @@
 public class Utils {
 
     public static void hideSoftKeyboard(Activity activity) {
-        InputMethodManager inputMethodManager =
-                (InputMethodManager) activity.getSystemService(
-                        Activity.INPUT_METHOD_SERVICE);
-        inputMethodManager.hideSoftInputFromWindow(
-                activity.getCurrentFocus().getWindowToken(), 0);
+        if (activity.getCurrentFocus() != null) {
+            InputMethodManager inputMethodManager =
```
jawidMuhammadi, line 20: Wouldn't it be much better if we use `hideSoftKeyboard(Activity activity)` method from `Utils` class?

dubeyaayush07: I am already using that method just modified it to check for null focus

jawidMuhammadi: Ohh! Yes you are, it is my mistake.

rachittt:
Thank you for your contribution!

PR #605, opened 2019-12-15
fix: layout on signup screen

akkywadhwa:
Fixes #571 
ezgif com-video-to-gif(5)

- [X] Apply the `AndroidStyle.xml` style template to your code in Android Studio.

- [X] Run the unit tests with `./gradlew check` to make sure you didn't break anything

- [X] If you have multiple commits please combine them into one commit by squashing them.

PR #861, opened 2020-03-28
Update the activity_mobile_verification

abisolaoye:
Centralising the editext OTP

Centralised the linearlayout of the OTP

fomerly this is the UI 
Screenshot_20200328-232013

After the changes in the UI

Screenshot_20200330-132422


Bhavnaharitsa:
@abisolaoye Please a relevant issue open to it. And since this is a UI change, add the screenshot too

abisolaoye:
Implemented the new icons just as we asked to do

formally it was like this

Screenshot_20200403-124213
Screenshot_20200403-124131


after the implementation of the icons in the code



Screenshot_20200404-180253
Screenshot_20200404-180234
Screenshot_20200404-180245

Bhavnaharitsa:
@abisolaoye The pr was based on the centralizing the buttons. Please make sure you make one change in one pr. 
And also, have you confirmed from anyone about the icons? Because they don't like the mifos one. Maybe I am missing out on the new ones.
Anyway, it's better to consult the mentors before this issue before asaik, the icons come first and then the text or they don't seem that necessary at all.

Bhavnaharitsa:
@abisolaoye Also, you have committed from the dev branch. Make sure, you open new branch for every pr you make :)

mifospay/src/main/res/layout/toolbar.xml
```diff
@@ -8,4 +8,10 @@
     app:contentInsetRight="0dp"
     app:contentInsetEnd="0dp"
     android:layout_width="match_parent"
-    android:layout_height="wrap_content" />
+    android:layout_height="wrap_content">
+    <ImageView
+        android:layout_width="@dimen/value_50dp"
+        android:layout_height="50dp"
```
jawidMuhammadi, line 14: Here also use the dynamic value from `dimen.xml` file which you are using for width. Please avoid hardcoded values
and also don't forget to squash your commits into one.

abisolaoye: Thank you

luckyman20:
@abisolaoye Sorry, but we have already similar PR #605 which is redesigning the whole signup screen. Also, I personally don't think logo besides title would look good but for community-discussion you can open an issue for it.

For future, I would recommend you to first open an issue for a bug or feature before sending a PR. Thank you for your contributions. :-)

PR #538, opened 2019-04-18
feat: Added Internet connection check in Login and Registration activities

RajaVamsi11:
Fixes #537 

Please Add Screenshots If any UI changes.

**Login Activity:**

Screenshot_2019-04-18-17-38-17

**Registration Activity:**

Screenshot_2019-04-18-18-44-06

Please make sure these boxes are checked before submitting your pull request - thanks!

- [x] Apply the `AndroidStyle.xml` style template to your code in Android Studio.

- [x] Run the unit tests with `./gradlew check` to make sure you didn't break anything

- [x] If you have multiple commits please combine them into one commit by squashing them.

Issue #252, opened 2019-01-26
No Email validator present in the Sign Up Details page. 


jddeep:
**Summary:**There is no validation check for correct email while filling up details.

**Steps to reproduce:**
Go to Sign up.
Proceed.
enter details.

**Expected behavior:**
To have a validation check on  the email,

**Observed behavior:**

What did you see instead?  Describe your issue in detail here.

**Device and Android version:**

What make and model device (e.g., Samsung Galaxy S3) did you encounter this on?  What Android
version (e.g., Android 4.0 Ice Cream Sandwich or Android 6.0 Marshmallow) are you running?  Is it
 the stock
version from the manufacturer or a custom ROM?

**Screenshots:**

jddeep:
@luckyman20 what do you think on this issue? I would like to work on this issue.

ankurs287:
@jddeep Email validation method is already defined in mifospay.utils package.

jddeep:
@ankurs287 But i did not see that implemented while i was Signing up.

ankurs287:
@jddeep then call the method defined in mifospay.utils.ValidationUtils class

luckyman20:
This issue can be taken up anyone now. Please comment here before you start working.

@s-ayush2903 tagging you since you asked on its closed PR.

s-ayush2903:
Is there any plan to write the new code in Kt? As I found some kt libs. in the gradle files. If it is so then, am I allowed to change the file for textUtils to Kotlin as it is very small and won't require much efforts. 
Contrarily, if we change our mind later to write that file in kt. then by that time there would've been a lot of java stuff.

So what should I do now:- Just call that required function at the appropriate place or even change that function file to kt. as well?

Please Elaborate/Suggest

s-ayush2903:
(Taking up this issue)

luckyman20:
@s-ayush2903 can you please open up another issue for converting the file into Kotlin?

s-ayush2903:
> @s-ayush2903 can you please open up another issue for converting the file into Kotlin?

#903 Has been opened considering the same

PR #913, opened 2020-05-18
fixes#881 - added splashscreen

### LINKED ISSUE #881: feat: Intro Splash Screen
**Summary:**
I think we need to have an intro screen since the logo of the app is finalised and also it would look good aesthetically.
The Intro Screen would come only when the app is installed for the first time or when the user kills the app.

devansh-299:
## Issue Fix
Fixes #881 

## Screen Recording
GIF-200518_194127

##

### DIFF
--- a/mifospay/src/main/AndroidManifest.xml
+++ b/mifospay/src/main/AndroidManifest.xml
@@ -17,16 +17,20 @@
         android:supportsRtl="true"
         android:theme="@style/AppTheme">
         <activity android:name=".merchants.ui.MerchantTransferActivity"></activity>
-        <activity
-            android:name=".auth.ui.LoginActivity"
-            android:theme="@style/AppTheme"
-            android:windowSoftInputMode="stateVisible">
+
+	<activity
+            android:name=".SplashScreenActivity"
+            android:theme="@style/SplashTheme">
             <intent-filter>
                 <action android:name="android.intent.action.MAIN" />
-
                 <category android:name="android.intent.category.LAUNCHER" />
             </intent-filter>
         </activity>
+        
+	<activity
+            android:name=".auth.ui.LoginActivity"
+            android:theme="@style/AppTheme"
+            android:windowSoftInputMode="stateVisible" />
         <activity
             android:name=".home.ui.MainActivity"
             android:windowSoftInputMode="adjustPan"
--- a/mifospay/src/main/java/org/mifos/mobilewallet/mifospay/SplashScreenActivity.java
+++ b/mifospay/src/main/java/org/mifos/mobilewallet/mifospay/SplashScreenActivity.java
@@ -1,0 +1,25 @@
+package org.mifos.mobilewallet.mifospay;
+
+import android.content.Intent;
+import android.content.pm.ActivityInfo;
+import android.os.Handler;
+import android.support.v7.app.AppCompatActivity;
+import android.os.Bundle;
+import org.mifos.mobilewallet.mifospay.auth.ui.LoginActivity;
+
+public class SplashScreenActivity extends AppCompatActivity {
+
+    private final int splashScreenDuration = 2000;
+
+    @Override
+    protected void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        setRequestedOrientation(ActivityInfo.SCREEN_ORIENTATION_PORTRAIT);
+        new Handler().postDelayed(new Runnable() {
+            public void run() {
+                startActivity(new Intent(SplashScreenActivity.this, LoginActivity.class));
+                finish();
+            }
+        }, splashScreenDuration);
+    }
+}
--- a/mifospay/src/main/res/drawable/splash_screen_background.xml
+++ b/mifospay/src/main/res/drawable/splash_screen_background.xml
@@ -1,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<layer-list xmlns:android="http://schemas.android.com/apk/res/android">
+
+    <item android:drawable="@android:color/white" />
+    <item
+        android:width="200dp"
+        android:height="200dp"
+        android:drawable="@drawable/mifospay_round_logo"
+        android:gravity="center" />
+
+</layer-list>
--- a/mifospay/src/main/res/values/styles.xml
+++ b/mifospay/src/main/res/values/styles.xml
@@ -1,4 +1,4 @@
-<resources>
+<resources xmlns:tools="http://schemas.android.com/tools">
 
     <style name="AppTheme" parent="Theme.MaterialComponents.Light.NoActionBar">
         <item name="colorPrimary">@color/colorPrimary</item>
@@ -35,6 +35,13 @@
 
     <style name="transparentBackground" >
         <item name="android:background">@null</item>
+    </style>
+
+    <!--SplashScreen Theme-->
+    <style name="SplashTheme" parent="Theme.AppCompat.NoActionBar">
+        <item name="android:statusBarColor"
+            tools:targetApi="lollipop">@color/colorPrimaryDark</item>
+        <item name="android:windowBackground">@drawable/splash_screen_background</item>
     </style>
 
 </resources>


mifospay/src/main/AndroidManifest.xml
```diff
@@ -17,16 +17,20 @@
...
             <intent-filter>
                 <action android:name="android.intent.action.MAIN" />
-
                 <category android:name="android.intent.category.LAUNCHER" />
             </intent-filter>
         </activity>
+        
+	<activity
+            android:name=".auth.ui.LoginActivity"
+            android:theme="@style/AppTheme"
```
Bhavnaharitsa, line 32: @devansh-299 Just popped out of my mind, did you have to add the entire LoginActivity again. I remember we having it. Perhaps the change only is enough. I might be wrong as well. Please correct me if I am :)

devansh-299: @Bhavnaharitsa I didn't add `LoginActivity` again, I just removed Launcher filter from this activity and used `SplashScreenActivity` as the new Launcher activity.

devansh-299:
@rachittt please have a look

mifospay/src/main/AndroidManifest.xml
Bhavnaharitsa: Yess, I agree. But the commit message shows as such :)

devansh-299: > Yess, I agree. But the commit message shows as such :)

yeah, the Github's change tracker is a little bit misleading

Bhavnaharitsa: Haha, same happened with me in the previous pr GitHub is surely misleading

shiv07tiwari:
@devansh-299 Maybe we can check the permissions and internet connection in the background while the user is looking into the splash screen. So isn't it better to use the Timer instead of using the theme and Cold Start time which can't be changed as per our requirements?
Your thoughts @luckyman20 @Bhavnaharitsa

s-ayush2903:
> @devansh-299 Maybe we can check the permissions and internet connection in the background while the user is looking into the splash screen. 

(This is just regarding the internet connection method)
Days back, I planned of implementing some networkChecker method in the module but surfing in the repo, I encountered [this review](https://github.com/openMF/mobile-wallet/pull/538#pullrequestreview-245891262) on the same idea. I was planning of implementing that method as some singleton(like those method which we have in Utils class) and using it in the module wherever required. 
So, is it needed/required now? I'll be eager to file a PR regarding the same.
Also, one more thing as discussed/pointed out previously in old discussions (like
* [This one](https://github.com/openMF/mobile-wallet/issues/252#issuecomment-628184426) and 
* [Second One](https://github.com/openMF/mobile-wallet/issues/252#issuecomment-628441270)), 

that Utils class again is quite small and easier to be ktfied. Is it allowed?

Bhavnaharitsa:
@shiv07tiwari Yes, setting up a timer is by far beneficial as it would help get us the required permissions the app, and like you said sir, the internet connectivities as well. 
@s-ayush2903 You mean, we'd be using or interfacing the singleton class wherever the internet connectivities are required? If that's the case, then we can use that in the Splash screen to check if the user has a valid connection to proceed to the SignIn process because right after the SplashScreen comes the SignIn page which I guess requires a good internet connection.

devansh-299:
> @devansh-299 Maybe we can check the permissions and internet connection in the background while the user is looking into the splash screen. So isn't it better to use the Timer instead of using the theme and Cold Start time which can't be changed as per our requirements?
> Your thoughts @luckyman20 @Bhavnaharitsa

@shiv07tiwari The basic application can run without having the read/write and camera 
permissions, so should we need necessarily need to ask for these permissions before hand?

devansh-299:
@shiv07tiwari @Bhavnaharitsa @s-ayush2903 I don't know whether this - introduction slides functionality is really something we need, but I once implemented this for Mifos Mobile [PR](https://github.com/openMF/mifos-mobile/pull/1271) where we can ask users for necessary permissions and give a basic introduction to the application.

s-ayush2903:
> @s-ayush2903 You mean, we'd be using or interfacing the singleton class wherever the internet connectivities are required? If that's the case, then we can use that in the Splash screen to check if the user has a valid connection to proceed to the SignIn process because right after the SplashScreen comes the SignIn page which I guess requires a good internet connection.

No, not interfacing, implementing like other methods(from Utils) are being used. 
I don't think checking internet connection at startup is a nice idea. Consider a scenario, where no internet connection is present, and app is launched. What happens then:- User gets notified regarding connection via some UI change right at the launch!, which decreases the UX right from the beginning of the launch, instead, the connection should be checked just before when a request is made that really needs connection rather checking the same beforehand.

Bhavnaharitsa:
> > @s-ayush2903 You mean, we'd be using or interfacing the singleton class wherever the internet connectivities are required? If that's the case, then we can use that in the Splash screen to check if the user has a valid connection to proceed to the SignIn process because right after the SplashScreen comes the SignIn page which I guess requires a good internet connection.
> 
> No, not interfacing, implementing like other methods(from Utils) are being used.
> I don't think checking internet connection at startup is a nice idea. Consider a scenario, where no internet connection is present, and app is launched. What happens then:- User gets notified regarding connection via some UI change right at the launch!, which decreases the UX right from the beginning of the launch, instead, the connection should be checked just before when a request is made that really needs connection rather checking the same beforehand.

@s-ayush2903 But what is needed is needed. The user needs an internet connection to sign in via google. But like you said, the internet connection varies from point to point. The user might have internet connection for now and suddenly something happens and he is out of internet. I personally suggested the permissions in the beginning for certain permissions like Camers which we have in QR code scanning and  certain other stuff namely doesnt come to my mind.

shiv07tiwari:
No adding a connectivity check only at the beginning isn't a good idea. I was just wondering if we might require some actions to be done alongside the splash screen, now or in the future. So isn't it better if we go with the timer method?

devansh-299:
> No adding a connectivity check only at the beginning isn't a good idea. I was just wondering if we might require some actions to be done alongside the splash screen, now or in the future. So isn't it better if we go with the timer method?

The timer approach will be better if we use it for actions like asking permissions or doing any other
background processing. So should I update the PR and add required alongside actions?

mifospay/src/main/java/org/mifos/mobilewallet/mifospay/SplashScreenActivity.java
```diff
@@ -0,0 +1,16 @@
+package org.mifos.mobilewallet.mifospay;
+
+import android.content.Intent;
+import android.support.v7.app.AppCompatActivity;
+import android.os.Bundle;
+import org.mifos.mobilewallet.mifospay.auth.ui.LoginActivity;
+
+public class SplashScreenActivity extends AppCompatActivity {
```
jawidMuhammadi, line 10: Splash screens are meant to engage users by the time app is launched and prevent showing a white screen in older devices while the OS starts the first activity and it should be very quick. If we don’t want to force our app users to wait when there is nothing to load -which is not recommended doing it, then there is no need to create a different activity for the `Splash Screen`. We can implement it with a lesser line of code. Please take a look at this [PR](https://github.com/apache/fineract-cn-mobile/pull/102) or go through this [article](https://android.jlelse.eu/launch-screen-in-android-the-right-way-aca7e8c31f52).

jawidMuhammadi (changes requested):

jawidMuhammadi:
> > No adding a connectivity check only at the beginning isn't a good idea. I was just wondering if we might require some actions to be done alongside the splash screen, now or in the future. So isn't it better if we go with the timer method?
> 
> The timer approach will be better if we use it for actions like asking permissions or doing any other
> background processing. So should I update the PR and add required alongside actions?

I think it is much better to ask permission whenever we need it, like when the user wants to open a camera , etc... so they will know why we are asking this permission. And what do you mean by background processing, like what?

mifospay/src/main/java/org/mifos/mobilewallet/mifospay/SplashScreenActivity.java
devansh-299: > Splash screens are meant to engage users by the time app is launched and prevent showing a white screen in older devices while the OS starts the first activity and it should be very quick. If we don’t want to force our app users to wait when there is nothing to load -which is not recommended doing it, then there is no need to create a different activity for the `Splash Screen`. We can implement it with a lesser line of code. Please take a look at this [PR](https://github.com/apache/fineract-cn-mobile/pull/102) or go through this [article](https://android.jlelse.eu/launch-screen-in-android-the-right-way-aca7e8c31f52).

Thanks you very much for the suggestion, I did read this article before. The current approach does handle the ColdStart's white screen but after going through other articles also, most of them suggested to use a separate activity. It does increase the code but it, in my opinion, this increases the code clarity. Thanks, if needed I shall update the PR :)

devansh-299:
> > > No adding a connectivity check only at the beginning isn't a good idea. I was just wondering if we might require some actions to be done alongside the splash screen, now or in the future. So isn't it better if we go with the timer method?
> > 
> > 
> > The timer approach will be better if we use it for actions like asking permissions or doing any other
> > background processing. So should I update the PR and add required alongside actions?
> 
> I think it is much better to ask permission whenever we need it, like when the user wants to open a camera , etc... so they will know why we are asking this permission. And what do you mean by background processing, like what?

I just proposed that if we use timer approach, then we can use that time for any required processing that we may need now or in the future.

mifospay/src/main/java/org/mifos/mobilewallet/mifospay/SplashScreenActivity.java
jawidMuhammadi: Ook, that is nice :). Actually I don't know why most of them suggest a separate activity if we don't need any background processing, although, activities are heavy and expensive operation for OS and it should be used only if we need an independent screen, as you know even `Google`'s recommended approach for app development is a `single activity` approach. So if you think it really necessary to create an entirely different activity for showing a logo to the screen because of increasing the code clarity then that is a valid point too :)

devansh-299:
@shiv07tiwari I have updated the PR as per discussion

shiv07tiwari (approved):
LGTM

PR #952, opened 2020-05-22
refactor: Converted Utils Class to Kotlin Object

### LINKED ISSUE #922: chore: Convert Utils Class to Kotlin Object
Summary:
Currently, the Utils class is written in java, it should be translated to kotlin instead which makes the code much more legible and cleaner.

Steps to reproduce:
Since it is a chore issue so it can directly be seen in the codebase at
`mobile-wallet/mifospay/src/main/java/org/mifos/mobilewallet/mifospay/utils`

Expected behavior:
It should be written in kotlin instead

Observed behavior:
The concerned class is written in java

Device and Android version:
NA

Screenshots:
NA

s-ayush2903:
## Issue Fix
Fixes #922 

## Screenshots

NA
## Description

Converted the concerned file to kt Object and made desired changes in the corresponding files
##


mifospay/src/main/java/org/mifos/mobilewallet/mifospay/auth/ui/LoginActivity.java
```diff
@@ -101,7 +101,7 @@ private void handleLoginInputChanged() {
 
     @OnClick(R.id.btn_login)
     public void onLoginClicked() {
-        Utils.hideSoftKeyboard(this);
+        Utils.INSTANCE.hideSoftKeyboard(this);
```
jawidMuhammadi: Do we really need `INSTANCE` here and in other files, can't we call the method directly like: `Utils.hideSoftKeyboard(this);`?

s-ayush2903:
As of now, No. One can't refer a non static method from a static context.
So each and every time, to use that, one'll have to create an instance of
same.
But, there must be some better way of it, rather repeating lines, lemme go
through kt docs, I'll update if possible.

s-ayush2903:
Maximum effort is made to keep this PR as semantic as possible(refraining from making changes in other files), so for that, methods have been annotated as static in the kt object file. Which works well and fine now, but as soon as the files(having implementation of those methods/using those methods) are in kt, then static annotations from those functions will also have to be removed(only to kill redundancy)

jawidMuhammadi (approved):
LGTM :)

s-ayush2903:
Please Review

s-ayush2903:
@rachittt @luckyman20 Please look into it

mifospay/src/main/java/org/mifos/mobilewallet/mifospay/utils/Utils.kt
```diff
@@ -0,0 +1,34 @@
...
+import android.view.inputmethod.InputMethodManager
+import java.text.NumberFormat
+import java.util.*
+
+object Utils {
+
+    @JvmStatic
+    fun hideSoftKeyboard(activity: Activity) {
+        val view = activity.currentFocus
+        val imm =
```
shiv07tiwari: Just rename this variable imm to it's elaborate camel case form. Rest looks good.

shiv07tiwari (changes requested):

shiv07tiwari (approved):

PR #854, opened 2020-03-23
feat: Added Passcode Verification for editing user credential

refactor-droidyy:
## Issue Fix
Fixes #853 

## Screenshots

ezgif com-gif-maker (2)

##

- [x] Apply the `AndroidStyle.xml` style template to your code in Android Studio.

- [x] Run the unit tests with `./gradlew check` to make sure you didn't break anything

- [x] If you have multiple commits please combine them into one commit by squashing them.

PR #684, opened 2020-01-27
fixes#683 - Passcode verification added

devansh-299:
Fixes #683 

Screen Recording

GIF-200127_192038

Please Add Summary of what changes you have made.(Optional)

Please make sure these boxes are checked before submitting your pull request - thanks!

- [x] Apply the `AndroidStyle.xml` style template to your code in Android Studio.

- [x] Run the unit tests with `./gradlew check` to make sure you didn't break anything

- [x] If you have multiple commits please combine them into one commit by squashing them.

PR #971, opened 2020-06-01
Add Passcode verification for editing user credential

### LINKED ISSUE #853: Add Feature for Passcode Verification to Edit Profile Credential
**Summary:**

Since the Edit profile section contains important data and if someone tries to edit profile user should again verify the identity.

refactor-droidyy:
## Issue Fix
Fixes #853 

## Screenshots

ezgif com-video-to-gif (5)

## Description

Added passcode verification for user editing user credentials.

##

### DIFF
--- a/mifospay/src/main/java/org/mifos/mobilewallet/mifospay/home/ui/ProfileFragment.java
+++ b/mifospay/src/main/java/org/mifos/mobilewallet/mifospay/home/ui/ProfileFragment.java
@@ -11,6 +11,8 @@ import android.view.ViewGroup;
 import android.widget.ImageView;
 import android.widget.TextView;
 
+import com.mifos.mobile.passcode.utils.PassCodeConstants;
+
 import org.mifos.mobilewallet.core.domain.model.client.Client;
 import org.mifos.mobilewallet.mifospay.R;
 import org.mifos.mobilewallet.mifospay.base.BaseActivity;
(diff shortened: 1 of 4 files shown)


refactor-droidyy:
@shiv07tiwari please review

mifospay/src/main/java/org/mifos/mobilewallet/mifospay/home/ui/ProfileFragment.java
```diff
@@ -135,6 +138,12 @@ public void onUserImageEditClicked() {
     public void onEditProfileClicked() {
         if (getActivity() != null) {
             getActivity().startActivity(new Intent(getActivity(), EditProfileActivity.class));
```
devansh-299: There's no need for this line as you are already creating an intent `PassCodeActivity` in the very next line

devansh-299 (changes requested):
@refactor-droidyy currently in your implementation there's a by-pass to this security check. Tap on User Profile Image, and you are in `EditProfileActivity` without any passcode verification.
Thank you for fixing the crashing issue as mentioned [here](https://github.com/openMF/mobile-wallet/pull/854#pullrequestreview-413368584) but I still don't think user should be directly logged out on failing. @shiv07tiwari @luckyman20 what are your views. Also "Forgot Password, Login Manually" issue still persists,to overcome these I think we may need to override the methods as done [here](https://github.com/openMF/mobile-wallet/pull/684), this is not the best way but its the only work around that I could find

mifospay/src/main/java/org/mifos/mobilewallet/mifospay/home/ui/ProfileFragment.java
refactor-droidyy: @devansh-299 I think this behaviour of user logging out if the passcode entered is incorrect 3 times it would be good for the user account security

luckyman20:
Yeah, I agree with @devansh-299

It could be something like if you enter wrong passcode for 3 times then you're logged out.

refactor-droidyy:
> Yeah, I agree with @devansh-299
> 
> It could be something like if you enter wrong passcode for 3 times then you're logged out.

@luckyman20 this PR has the above implementation that if you enter wrong passcode for 3 times then you're logged out.

luckyman20:
In that case, @devansh-299 do you have anything else in mind which might suit better to this use case?

devansh-299:
> In that case, @devansh-299 do you have anything else in mind which might suit better to this use case?

No, I was just doubtful whether this functionality should be added or not.

therajanmaurya:
Unrelated and not needed anymore

PR #973, opened 2020-06-11
Feat: Added support for Standing Instructions

devansh-299:
## Description
Added complete support for Standing Instructions. Users can now create new standing instructions, get client specific standing instructions and update/delete standing instructions created earlier.
**Note:** Currently the update and create functionality are not working because of Fineract issues

### Creating New Standing Instruction
GIF-200611_213537
---

### Existing Standing Instruction
GIF-200611_214251

---

## Screen Shots 
|||||

---

### DIFF
--- a/core/src/main/java/org/mifos/mobilewallet/core/data/fineract/entity/payload/StandingInstructionPayload.kt
+++ b/core/src/main/java/org/mifos/mobilewallet/core/data/fineract/entity/payload/StandingInstructionPayload.kt
@@ -1,0 +1,18 @@
+package org.mifos.mobilewallet.core.data.fineract.entity.payload
+
+import android.os.Parcelable
+import kotlinx.android.parcel.Parcelize
+
+@Parcelize
+data class StandingInstructionPayload(var fromOfficeId : Int, var fromClientId : Int,
+                                      var fromAccountType : Int, val name : String?,
+                                      val transferType : Int, val priority : Int,
+                                      val status : Int, var fromAccountId : Long,
+                                      var toOfficeId : Int, var toClientId : Int,
+                                      var toAccountType : Int, var toAccountId : Long,
+                                      val instructionType : Int, var amount : Double,
+                                      var validFrom : String?, val recurrenceType : Int,
+                                      val recurrenceInterval : Int, val recurrenceFrequency : Int,
+                                      val locale : String?, val dateFormat : String?,
+                                      var validTill : String?, var recurrenceOnMonthDay : String?,
+                                      val monthDayFormat : String?) : Parcelable
--- a/core/src/main/java/org/mifos/mobilewallet/core/domain/usecase/standinginstruction/CreateStandingTransaction.kt
+++ b/core/src/main/java/org/mifos/mobilewallet/core/domain/usecase/standinginstruction/CreateStandingTransaction.kt
@@ -1,0 +1,191 @@
+package org.mifos.mobilewallet.core.domain.usecase.standinginstruction
+
+import org.mifos.mobilewallet.core.base.UseCase
+import org.mifos.mobilewallet.core.data.fineract.entity.accounts.savings.SavingAccount
+import org.mifos.mobilewallet.core.data.fineract.entity.client.Client
+import org.mifos.mobilewallet.core.data.fineract.entity.client.ClientAccounts
+import org.mifos.mobilewallet.core.data.fineract.entity.standinginstruction.SDIResponse
+import org.mifos.mobilewallet.core.data.fineract.entity.payload.StandingInstructionPayload
+import org.mifos.mobilewallet.core.data.fineract.repository.FineractRepository
+import org.mifos.mobilewallet.core.utils.Constants
+import rx.Subscriber
+import rx.android.schedulers.AndroidSchedulers
+import rx.schedulers.Schedulers
+import javax.inject.Inject
+
+
+class CreateStandingTransaction @Inject constructor(private val apiRepository: FineractRepository) :
+        UseCase<CreateStandingTransaction.RequestValues, CreateStandingTransaction.ResponseValue>() {
+
+    lateinit var fromClient: Client
+    lateinit var toClient: Client
+    lateinit var fromAccount: SavingAccount
+    lateinit var toAccount: SavingAccount
+
+    override fun executeUseCase(requestValues: RequestValues) {
+        apiRepository.getSelfClientDetails(requestValues.fromClientId)
+                .observeOn(AndroidSchedulers.mainThread())
+                .subscribeOn(Schedulers.io())
+                .subscribe(object : Subscriber<Client>() {
+
+                    override fun onCompleted() {
+
+                    }
+
+                    override fun onError(e: Throwable) {
+                        useCaseCallback.onError(Constants.ERROR_FETCHING_CLIENT_DATA)
+                    }
+
+                    override fun onNext(client: Client) {
+                        fromClient = client
+                        fetchToClientData()
+                    }
+                })
+    }
+
+    private fun fetchToClientData() {
+        apiRepository.getClientDetails(requestValues.toClientId)
+                .observeOn(AndroidSchedulers.mainThread())
+                .subscribeOn(Schedulers.io())
+                .subscribe(object : Subscriber<Client>() {
+
+                    override fun onCompleted() {
+
+                    }
+
+                    override fun onError(e: Throwable) {
+                        useCaseCallback.onError(Constants.ERROR_FETCHING_CLIENT_DATA)
+                    }
+
+                    override fun onNext(client: Client) {
+                        toClient = client
+                        fetchFromAccountDetails()
+                    }
+                })
+    }
+
+    private fun fetchFromAccountDetails() {
+        apiRepository.getSelfAccounts(requestValues.fromClientId)
+                .observeOn(AndroidSchedulers.mainThread())
+                .subscribeOn(Schedulers.io())
+                .subscribe(object : Subscriber<ClientAccounts>() {
+                    override fun onCompleted() {
+
+                    }
+
+                    override fun onError(e: Throwable) {
+                        useCaseCallback.onError(Constants.ERROR_FETCHING_FROM_ACCOUNT)
+                    }
+
+                    override fun onNext(clientAccounts: ClientAccounts) {
+                        val accounts = clientAccounts.savingsAccounts
+                        if (accounts != null && accounts.size != 0) {
+                            var walletAccount: SavingAccount? = null
+                            for (account in accounts) {
+                                if (account.productId ==
+                                        Constants.WALLET_ACCOUNT_SAVINGS_PRODUCT_ID) {
+                                    walletAccount = account
+                                    break
+                                }
+                            }
+                            walletAccount?.let{
+                                fromAccount = walletAccount
+                                fetchToAccountDetails()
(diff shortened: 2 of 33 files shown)


core/src/main/java/org/mifos/mobilewallet/core/data/fineract/entity/payload/StandingInstructionPayload.kt
```diff
@@ -0,0 +1,85 @@
+package org.mifos.mobilewallet.core.data.fineract.entity.payload
+
+import android.os.Parcel
+import android.os.Parcelable
+
```
jawidMuhammadi, line 5: Instead of overriding all these methods from `Parcelable`, you can simply Add `@Parcelize` annotation here (on top of `data class  StandingInstructionPayload`) then you can simply delete all overridden methods. This [article](https://medium.com/@BladeCoder/a-study-of-the-parcelize-feature-from-kotlin-android-extensions-59a5adcd5909) might help if you want to learn how it works.

jawidMuhammadi (changes requested):

mifospay/src/main/res/layout/activity_si_details.xml
```diff
@@ -0,0 +1,280 @@
...
+                        android:textColor="@color/colorAccent"
+                        android:textSize="@dimen/value_16sp"
+                        android:layout_marginBottom="@dimen/marginItemsInSectionMedium" />
+
+                    <TextView
+                        android:layout_width="wrap_content"
+                        android:layout_height="wrap_content"
+                        android:textColor="@color/gray_dark"
+                        android:textSize="@dimen/value_16sp"
+                        android:text="Amount:"/>
```
jawidMuhammadi: It would be much better if it wasn't hardcoded otherwise will not be able to translate it :)

mifospay/src/main/res/layout/activity_new_si.xml
```diff
@@ -0,0 +1,208 @@
...
+                android:layout_height="@dimen/value_50dp"
+                android:text="SELECT DATE"
+                android:layout_gravity="center"
+                android:layout_marginBottom="@dimen/value_20dp"
+                android:textColor="@color/white" />
+            <Button
+                android:id="@+id/btn_create_si"
+                android:layout_width="260dp"
+                android:layout_height="@dimen/value_50dp"
+                android:text="Submit"
```
jawidMuhammadi: same here

mifospay/src/main/java/org/mifos/mobilewallet/mifospay/standinginstruction/ui/SIDetailsActivity.kt
```diff
@@ -0,0 +1,264 @@
...
+
+        val res = resources
+        iv_empty_no_transaction_history
+                    .setImageDrawable(res.getDrawable(drawable))
+        tv_empty_no_transaction_history_title.text = res.getString(errorTitle)
+        tv_empty_no_transaction_history_subtitle.text = res.getString(errorMessage)
+    }
+
+    override fun siDeletedSuccessfully() {
+        showToast("Deleted Successfully")
```
jawidMuhammadi: Same here

mifospay/src/main/java/org/mifos/mobilewallet/mifospay/standinginstruction/ui/SIDetailsActivity.kt
```diff
@@ -0,0 +1,264 @@
...
+        tv_si_from_name.text = "Name: ${standingInstruction.fromClient.displayName}"
+        tv_si_from_number.text = "Number: ${standingInstruction.fromAccount.accountNo}"
+        tv_si_to_name.text = "Name: ${standingInstruction.toClient.displayName}"
+        tv_si_to_number.text = "Number: ${standingInstruction.toAccount.accountNo}"
+
+        tv_si_status.text = standingInstruction.status.value
+        tv_recurrence_interval.text = standingInstruction.recurrenceInterval.toString()
+
+        // setting up TextInputLayouts
+        til_si_edit_amount.hint = "₹ ${standingInstruction.amount}"
```
jawidMuhammadi: Are we sure this amount will be in Rupee all the time? I think the symbol should be changed based on amount's currency dynamically.

mifospay/src/main/java/org/mifos/mobilewallet/mifospay/standinginstruction/ui/SIDetailsActivity.kt
```diff
@@ -0,0 +1,264 @@
...
+
+        tv_valid_from.text = "${standingInstruction.validFrom[2]}-" +
+                "${standingInstruction.validFrom[1]}-${standingInstruction.validFrom[0]}"
+        tv_valid_till.text = "${standingInstruction.validTill?.get(2)}-" +
+                "${standingInstruction.validTill?.get(1)}-${standingInstruction.validTill?.get(0)}"
+
+        tv_si_from_name.text = "Name: ${standingInstruction.fromClient.displayName}"
+        tv_si_from_number.text = "Number: ${standingInstruction.fromAccount.accountNo}"
+        tv_si_to_name.text = "Name: ${standingInstruction.toClient.displayName}"
+        tv_si_to_number.text = "Number: ${standingInstruction.toAccount.accountNo}"
```
jawidMuhammadi: Can we use `formatted string` dynamically in here? Like `<string name="name">Name: %1$s</string>`

jawidMuhammadi (changes requested):

core/src/main/java/org/mifos/mobilewallet/core/domain/usecase/standinginstruction/CreateStandingTransaction.kt
```diff
@@ -0,0 +1,191 @@
...
+                                if (account.productId ==
+                                        Constants.WALLET_ACCOUNT_SAVINGS_PRODUCT_ID) {
+                                    walletAccount = account
+                                    break
+                                }
+                            }
+                            walletAccount?.let{
+                                toAccount = walletAccount
+                                createNewStandingInstruction()
+                            }?: useCaseCallback.onError(Constants.NO_WALLET_FOUND)
```
jawidMuhammadi, line 127: `Constants.NO_WALLET_FOUND` do we show this error message in UI? If yes then it would be much better to add inside `string.xml` file

mifospay/src/main/java/org/mifos/mobilewallet/mifospay/standinginstruction/adapter/StandingInstructionAdapter.kt
```diff
@@ -0,0 +1,64 @@
...
+
+    class ViewHolder(itemView: View) : RecyclerView.ViewHolder(itemView) {
+
+        fun bindItems(standingInstruction: StandingInstruction) {
+            val tvFromClientName = itemView.findViewById(R.id.tv_from_client_name) as TextView
+            val tvToClientName = itemView.findViewById(R.id.tv_to_client_name) as TextView
+            val tvValidTill  = itemView.findViewById(R.id.tv_valid_till_date) as TextView
+            val tvAmount  = itemView.findViewById(R.id.tv_amount) as TextView
+
+            tvFromClientName.text = "From: ${standingInstruction.fromClient.displayName}"
```
jawidMuhammadi: Use formatted string, please.

mifospay/src/main/java/org/mifos/mobilewallet/mifospay/standinginstruction/adapter/StandingInstructionAdapter.kt
```diff
@@ -0,0 +1,64 @@
...
+            tvToClientName.text = "To: ${standingInstruction.toClient.displayName}"
+            /**
+             * Using hardcoded Currency as response doesn't return the currency
+             */
+            tvAmount.text = "${Constants.RUPEE} ${standingInstruction.amount.toString()}"
+            val validTill  = " ${standingInstruction.validTill?.get(2)}" +
+                    "-${standingInstruction.validTill?.get(1)}" +
+                    "-${standingInstruction.validTill?.get(0)}"
+            tvValidTill.text = "Valid Till: $validTill"
```
jawidMuhammadi: same here

jawidMuhammadi (changes requested):

core/src/main/java/org/mifos/mobilewallet/core/domain/usecase/standinginstruction/CreateStandingTransaction.kt
```diff
@@ -0,0 +1,191 @@
...
+
+                        } else {
+                            useCaseCallback.onError(Constants.ERROR_FETCHING_TO_ACCOUNT)
+                        }
+                    }
+                })
+    }
+
+    private fun createNewStandingInstruction() {
```
s-ayush2903, line 136: @devansh-299 It would be much better if we use factory for initializing/creating new standingInstructions throughout the codebase.
Why?
* As it would reduce the following kind/type of repititive code in the concerned file(s) where it is needed to create a new standingInstruct'n

devansh-299: @s-ayush2903 I get what that factory pattern will reduce the boilerplate code but we just need this code at two places, hence creating a factory for this in my opinion will be an overkill

core/src/main/java/org/mifos/mobilewallet/core/domain/usecase/standinginstruction/CreateStandingTransaction.kt
devansh-299: Thanks for pointing this out

mifospay/src/main/java/org/mifos/mobilewallet/mifospay/standinginstruction/ui/SIDetailsActivity.kt
devansh-299: I forgot to mention in comments here, but the response doesn't specify the currency so for the time being, we are going with the hardcoded currency

core/src/main/java/org/mifos/mobilewallet/core/domain/usecase/standinginstruction/CreateStandingTransaction.kt
s-ayush2903: I'm astonished at consideration of factory as an _overkill_.Let me provide a few more instances which clicked right now(while authoring this thread)where this can be helpful
* Consider the scenario where one is trying to test this feature via api calls first;
* Then as for instruction creation takes 3 mandatory parameters, out of which vpa needs to be validated, so for end-to-end testing(covering all edge cases) one'll have to create two `SIs`-
a valid one and an invalid one,
* If we consider the unit test it solely(not networking calls, just its functionality), that'll also require same two more instances of the same
* So considering this minimum applicability/usages of SIs, the total locs would be=
  `41x(2+2+2)=246` {41=179-136} 
If we had factory then, it would be:- `41x1 + 4+4+4 = 53` {4=2x(1+1)}

But still, i was just putting forward the standard software design pattern and my invaluable view regarding the same, everything is upto you

devansh-299: @s-ayush2903 thank you providing such an in depth analysis and for staging your views. I will consult my mentors over this but anyways I appreciate your thoughtfulness

shiv07tiwari (approved):

PR #1025, opened 2020-08-04
Fixed keyboard

MaskedCarrot:
Closed keyboard while switching between mobile and vpa

##


s-ayush2903:
There is no need to make the method flavor specific,, you can directly use the method that you were trying to write from the extension functions that we have in utils class

s-ayush2903:
Also, this is not a very good practice on not checking the build on your local machine and pushing it here, also falsedly checkmarking the `./gradlew check` option in the PR description. As i can see, you just wrote that code directly by editing the file in github not in AS, please refrain this until you're making a documentation change :)

PR #1038, opened 2020-10-24
fix: VPA String validation

### LINKED ISSUE #1037: Not checking for valid VPA string in SendFragment
**Summary:**

In SendFragment there is no logic for VPA String pre validation . When user clicks on submit button, the app should pre validate the VPA String before sending it to SearchClient or before searching it in FineractRepository. Should be checked if the input is a valid VPA (and follows proper VPA formating) or it is just a random string. Using **Regular Expressions** to pre validate will save unnecessary API/Repository burden, It will make validation faster, and will be good for user experience.

**Steps to reproduce:**

1. Go to Payments from Home
2. Type Amount (less than TOTAL BALANCE)
3. Type any RANDOM stirng which can not be a valid VPA ID (eg. adkdjckjc, cjbdcjkancjd)
4. Click on SUBMIT button

**Expected behavior:**

Should first pre validate the VPA string with regular expressions and if it is found to be a valid vpa then only be searched in repository.
The same is done in Google pay app also to insntantly validate VPA as shown in the below screenrecording.

gpayss
_(screenRecording from gpay app)_

**Observed behavior:**

Directly uses any random input String to be searched in repository without prevalidation.

Honor 7x  
Android 9 Pie 
StockRom

**ScreenRecording:**

ezgif com-gif-maker

Me-Abhishek-patel:
## Issue Fix
Fixes #1037

## ScreenRecording
ezgif com-gif-maker (1)

## Description
Added VPA string validation with Regex.
Now App shows a snackbar ( Please enter a valid VPA) **Instantly** whenever user tries to input an invalid vpa.
##

### DIFF
--- a/mifospay/src/main/java/org/mifos/mobilewallet/mifospay/utils/Constants.java
+++ b/mifospay/src/main/java/org/mifos/mobilewallet/mifospay/utils/Constants.java
@@ -73,6 +73,7 @@ public class Constants {
     public static final String ERROR_FINDING_VPA = "Error finding Virtual Payment Address";
     public static final String ERROR_FINDING_MOBILE_NUMBER = "Error finding Mobile Number";
     public static final String PLEASE_ENTER_VALID_AMOUNT = "Please enter a valid amount";
+    public static final String VPA_VALIDATION_REGEX = "^\\w.+@\\w+$";
     public static final String SELF_ACCOUNT_ERROR = "Self Account transfer is not allowed";
     public static final String PLEASE_ENTER_AMOUNT =
             "Please enter a valid amount before making the transfer";
--- a/mifospay/src/main/java/org/mifos/mobilewallet/mifospay/payments/ui/SendFragment.java
+++ b/mifospay/src/main/java/org/mifos/mobilewallet/mifospay/payments/ui/SendFragment.java
@@ -146,6 +146,10 @@ public class SendFragment extends BaseFragment implements BaseHomeContract.Trans
                 showSnackbar(Constants.PLEASE_ENTER_VALID_AMOUNT);
                 return;
             }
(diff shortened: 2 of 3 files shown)


Me-Abhishek-patel:
@devansh-299 Please review this PR.

mifospay/src/main/java/org/mifos/mobilewallet/mifospay/utils/Constants.java
```diff
@@ -73,6 +73,8 @@
     public static final String ERROR_FINDING_VPA = "Error finding Virtual Payment Address";
     public static final String ERROR_FINDING_MOBILE_NUMBER = "Error finding Mobile Number";
     public static final String PLEASE_ENTER_VALID_AMOUNT = "Please enter a valid amount";
+    public static final String PLEASE_ENTER_VALID_VPA = "Please enter a valid VPA";
```
devansh-299: Instead of declaring inside `Constants.java` please create a _string resource_. Since this message will be displayed to the user, so if in the future we plan to provide multi-language support then this approach will not work.

devansh-299 (changes requested):

mifospay/src/main/java/org/mifos/mobilewallet/mifospay/utils/Constants.java
Me-Abhishek-patel: @devansh-299 I've updated my PR . Please review again.

devansh-299 (approved):
Seems good to me!

devansh-299:
@Me-Abhishek-patel can you squash the commits into a single commit?

Me-Abhishek-patel:
> @Me-Abhishek-patel can you squash the commits into a single commit?

thank you @devansh-299, I have updated my PR.

PR #1044, opened 2020-11-09
CI: Migrated CI from Travis to GitHub Actions

### LINKED ISSUE #1043: Add CI/CD GitHub Actions and remove other build checks
**Summary:**

The project currently has Codeship, CircleCI, Travis We need to remove these and use Github actions.
Following are the requirements:-

* Generate the apk for every pull request and every commit on the dev and master branch.
* Build checks for every commit on any other branch.

s-ayush2903:
## Issue Fix
Fixes #1043 

## Screenshots
NA

## Description
Added workflow file and removed previous options

##


s-ayush2903:
The artifacts are being generated and uploaded on pushes. But I'm not sure why the CI checks ain't being triggered here as I opened the same PR at my dev with exactly same changes and the CI there was triggered, could be some glitch, maybe another PR would not show the same problem. Anyways, the corresponding push run: [here](https://github.com/s-ayush2903/mobile-wallet/runs/1377235255). The PR of which I was talking: [here](https://github.com/s-ayush2903/mobile-wallet/pull/1/)

shiv07tiwari (changes requested):
@s-ayush2903  Nice work. But there are some things still missing in the file. We need to do a simple build check on every feature branch (all except dev and master) and we need build check + apk for every PR and every commit on dev and master.
You can refer to the following article: https://wkrzywiec.medium.com/github-actions-for-android-first-approach-f616c24aa0f9

shiv07tiwari:
@s-ayush2903  Please add this to the README as well after removing the Travis section.

s-ayush2903:
Fine, I'll definitely make the required changes.

s-ayush2903:
I'm not sure why checks ain't being triggered here at this PR whereas they are working at my PR on my local repo. I'll check if a close and reopen of this PR produces some effect. Or if doesn't then I'll open another PR and then see.

shiv07tiwari:
@s-ayush2903 The actions won't be triggered as the repo is not yet configured for the same. Once we merge your changes, the actions should trigger. Just make the changes using 2 separate yml files for feature branch and master/dev.

s-ayush2903:
Okay fine then. I was wondering that they are being triggered at my fork but not here. Now got it!

.github/workflows/feature_branch_ci.yml
```diff
@@ -0,0 +1,26 @@
+name: CI Check for Mobile-Wallet
```
shiv07tiwari: Rename this to Mobile-Wallet Feature Branch CI. It shall refect that it is for the features part.

.github/workflows/feature_branch_ci.yml
```diff
@@ -0,0 +1,26 @@
...
+on:
+  push:
+    branches:
+      - '*'
+      - '!dev'
+      - '!master'
+jobs:
+  build:
```
shiv07tiwari, line 11: Add name alongside runs-on for this jobId

.github/workflows/main_branches_ci.yml
```diff
@@ -0,0 +1,32 @@
+name: CI Check for Mobile-Wallet
```
shiv07tiwari: Same as the last file, rename this. Also IMO we should rename this file to something like master_dev_ci.

.github/workflows/pull_request_ci.yml
```diff
@@ -0,0 +1,28 @@
+name: CI Check for Mobile-Wallet
```
shiv07tiwari: This is exactly the same except for when to trigger the action, right? Just add the pull_request alongside push in the last file. 
on:
  pull_request:
    branches:
      - 'master'
  push:
    branches:
      - 'master'

build.gradle
```diff
@@ -35,6 +35,7 @@ ext {
     compileSdkVersion = 28
     buildToolsVersion = '26.0.2'
     kotlinVersion = "1.3.61"
+    ndkVersion = '21.3.6528147'
```
shiv07tiwari, line 38: Why have you added this line?

.github/workflows/main_branches_ci.yml
s-ayush2903: Okay, what do you think about the name: `Mobile Wallet CI[Master/Dev]`

build.gradle
s-ayush2903: Actually github actions require ndk version to be specified for the check to run. It is a known issue, not fixing it causes the execution for `stripDebugDebugSymbols` fail which in turn fails entire process. Official Docs: [here](https://developer.android.com/studio/projects/install-ndk#apply-specific-version); known issue: [here](https://github.community/t/android-ndk-on-unix-is-incompatible-with-android-build-tools-3-6-0/17697)

.github/workflows/master_dev_ci.yml
```diff
@@ -0,0 +1,38 @@
+name: Mobile-Wallet CI[Master/Dev-Branch]
+on:
+  push:
+    branches:
+      - 'dev'
+      - 'master'
+
+  pull_request:
+    branches:
+      - '*'
```
shiv07tiwari: We don't need to add all the branches here. I think adding for dev and master PRs is suffiecient. Just change this and I think we are good to go. :smile:

shiv07tiwari (changes requested):

.github/workflows/master_dev_ci.yml
s-ayush2903: But what about the hotfix branches then, should we consider build on PRs targeting 'em?

.github/workflows/main_branches_ci.yml
shiv07tiwari: Yes sure.

.github/workflows/master_dev_ci.yml
shiv07tiwari: We can add them as per the necessity, but adding for all the branches is not required. I think we can add payment-hub as well here. Add payment-hub in PR and push.

s-ayush2903: Fine!

shiv07tiwari (approved):

PR #1049, opened 2020-11-11
Fix: Fix syntax in yaml file & Upload Checkstyle Report to GH-Actions

### LINKED ISSUE #1048: Invalid syntax in yaml file for Github actions
@s-ayush2903 The YAML file for Github actions give syntax error at line 12. The name should go inside the build job. Kindly fix this. 
jobs:
  build:
    name: Build APK
    runs-on: ubuntu-latest

Ref: https://github.com/openMF/mobile-wallet/pull/1044

s-ayush2903:
## Issue Fix
Fixes #1048 

## Screenshots
NA

## Changes Made
- [x] Fix Syntax issue at line 12
- [x] Upload Codestyle check reports to GH-Actions(on failure only)
- [x] Reverted changes in gradle files(from #1044)
- [x] Specified NDK in CI flow only

##


shiv07tiwari:
@s-ayush2903 Line no 11 has the syntax error. Add a tab to fix the indentation.

shiv07tiwari (approved):

s-ayush2903:
@shiv07tiwari It is my very humble request to you to please not tell me how to fix the error. Please just mention there is an error maybe at this file. As by telling me where error is and what it is and its fix won't lead me to learn niches and troubleshooting. Thanks for help though!

devansh-299 (changes requested):
@s-ayush2903 can you also add a job for checkstyles, we already have a config file for that (see [here](https://github.com/openMF/mobile-wallet/blob/dev/config/quality/checkstyle/checkstyle-config.xml)) and we can make use of it while reviewing pull requests

s-ayush2903:
Ah, a nice idea indeed! So for Checkstyle what are your thoughts if we generate the report as well(as an artifact) after the check has finished, it can become easier for new contributors to find where they fell short of checkstyle.

devansh-299:
> 
> 
> Ah, a nice idea indeed! So for Checkstyle what are your thoughts if we generate the report as well(as an artifact) after the check has finished, it can become easier for new contributors to find where they fell short of checkstyle.

Yeah that will be great!

s-ayush2903:
@shiv07tiwari Sir, should we upload reports all the time(means even in the cases where they pass the checks) or only in the failures. Uploading reports in all cases would make them believe that this uploading works really and they would be able to see the reports how they look when everything is right&clean(when all checks passes), if we upload reports on failures only then people will always know how the report corresponding to failure looks like not for the sucess scenarios. Anyways, this was my ideology, I'm leaving this up to you to decide when to upload the reports.

As of now, the reports are being uploaded in both the cases.

devansh-299 (approved):
Looks good me! 
For that report generation, I think creating them in case of any failures makes more sense

s-ayush2903:
Cool then, no worries :v:

shiv07tiwari (changes requested):
@s-ayush2903 I just wanted to ask, why have you removed the Dockerfile?

s-ayush2903:
I found no use of it, because we don't need it. If we plan releases then we can do 'em by GH-Actions. I can revert the changes if we can make use of the concerned file.

shiv07tiwari:
We don't need to remove that file. We should Dockerise the project. Soon we are planning to open an issue to create the docker-compose file for the same. Kindly add the file. Thanks :)

s-ayush2903:
Oh niceee, fine!

s-ayush2903:
One more thing I have to say is that I have figured out a way to avoid specifying ndkVersion in the gradle file and make build at ci pass without it. We can revert those changes and install the NDK only in CI process only when needed, no need to change any gradle files : ) I have tested it [here](https://github.com/s-ayush2903/mobile-wallet/pull/2) and it works fine. 

Seeking permission to revert changes in gradle files as well(that were introduced in #1044).

shiv07tiwari:
@s-ayush2903 Looks good. Do the required changes.

shiv07tiwari (approved):
Good Job @s-ayush2903

s-ayush2903:
I think we should update CI badges as well(present in the readme), currently they show last travis ci status. It can be a 
`good-first-issue` for the people wishing to make their initial contributions to the project.
|Outdated Badges[CI]|
|--|
|travisBadges|

s-ayush2903:
Created #1050 :+1:

PR #1071, opened 2020-12-06
fix : converted all inner class based alert dialog box to our global class based dialog box 

### LINKED ISSUE #660: fix: Convert inner class of Alert Dialog box to our class DialogBox
**Summary:**
Our app still has some alert boxes extracted from the inner class in Android. But, since we have our own global alert box, I think all the alert boxes that was previously built using inner class can be converted.

ReCodee:
## Issue Fix
Fixes #660 

## Description

Changes made:

- Converted all android inner class based alert dialog boxes to our own global class based dialog box.

- Added `NeutralListener` in `DialogBox.java` to provide functionality to add `NeutralButton` in the DialogBox.

- Added a parameter for `View` in `show()` to provide an option to set a view in the dialog box or null can be passed.

- Overloaded `show()` methods in `DialogBox.java` to facilitate different types of dialog boxes for eg: with or without a message and with or without a neutral button. 

##


ReCodee:
@devansh-299 PTAL :)

PR #1096, opened 2021-01-03
Fix#1093

### LINKED ISSUE #1093: UI enhancement in activity_read_qr.xml
**Summary:**

The UI in activity_read_qr.xml is a bit odd. The activity is used for scanning QR code.

**Steps to reproduce:**

The toolbar on top should be removed and the back button and scan code should be at a transparent frame.

**Expected behavior:**

The scan code activity should be transparent and every UI component should appear on that transparent frame

**Observed behavior:**

The issue with the UI of this activity is that it contains a toolbar on top which is a bit outdated and the flash on/off ImageView should also be at top and small in size.

VIVO Y83, android version - 8.1.0

**Screenshots:**

EGOR-IND:
## Issue Fix
Fixes #1093 

## Description
I updated the UI of QR scanner and put some margin between history tab tags and and some attributes that are supported by API level less that 17 because there were some attributes in some layout files which are not supported in API level less than 17. Sorry for not squashing commits but actually in these 3 commits one is add commit and other is reverting that add so it's actually one commit, I'm learning, sorry for inconvenience.

##


EGOR-IND:
@shiv07tiwari @devansh-299 @robustTechie Please check this PR.

PR #1151, opened 2021-01-10
Fix : #1088 Adding Toasts for perticular edittext field

### LINKED ISSUE #1088: Feat: Adding Toast for  Amount, Virtual Payment Address and Mobile Number in Payments.
**Summary:**

Particular empty edittext field is not mentioned by toast on pressing submit button.

**Steps to reproduce:**

1. Go to Payments.
2. Go to SEND fragment.
3. Inside SEND fragment tap on VPA Chip or Mobile Number chip.
4. There you can press the submit button without entering anything inside both (or any) of the edittext.

**Expected behavior:**

After pressing the submit button if any of the edittext fields is empty the user should be highlighted edittext field which is empty. 

**Observed behavior:**
Application is just displaying a general toast("Please enter all the fields") not specifying which field, in particular, is not filled.

Redmi Note 5 Pro
Android 9 Pie
custom ROM

**Screenshots:**

WhatsApp Image 2020-12-23 at 11 23 32 AM
WhatsApp Image 2020-12-23 at 11 23 31 AM
WhatsApp Image 2020-12-23 at 11 23 31 AM (1)

Harshad09:
## Issue Fix
Fixes #1088

## Screenshots

WhatsApp Image 2021-01-10 at 1 21 57 PM (2)
WhatsApp Image 2021-01-10 at 1 21 57 PM (1)
WhatsApp Image 2021-01-10 at 1 21 57 PM

## Description

Added Toast for particular edittext if it is kept empty in sendFragment in payments.

##

### DIFF
--- a/mifospay/src/main/java/org/mifos/mobilewallet/mifospay/payments/ui/SendFragment.java
+++ b/mifospay/src/main/java/org/mifos/mobilewallet/mifospay/payments/ui/SendFragment.java
@@ -136,10 +136,20 @@ public class SendFragment extends BaseFragment implements BaseHomeContract.Trans
         String eamount = etAmount.getText().toString().trim();
         String mobileNumber = mEtMobileNumber.getText()
                 .toString().trim().replaceAll("\\s+", "");
-        if (eamount.equals("") || (mBtnVpa.isSelected() && externalId.equals("")) ||
-                (mBtnMobile.isSelected() && mobileNumber.equals(""))) {
+        if (eamount.equals("")) {
             Toast.makeText(getActivity(),
-                    Constants.PLEASE_ENTER_ALL_THE_FIELDS, Toast.LENGTH_SHORT).show();
+                    Constants.EMPTY_AMOUNT, Toast.LENGTH_SHORT).show();
+            return;
+        }
+        if (externalId.equals("") && mBtnVpa.isSelected()) {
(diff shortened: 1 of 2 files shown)


Harshad09:
@devansh-299  @s-ayush2903 can you please review my pull request.

devansh-299:
@Harshad09 first of all I don't it's required to show separate toasts for all the fields as we only have 3 fields on the screen and user can very well understand which filled needs to be filled when the `Please enter all fields` message is displayed. This would have made sense in case we were having a long-form where the user may want to know what exactly was left incomplete.

Also, your code will fail for cases like when both `amount` and `externalId` are empty, etc. For handling these a lot of boilerplate will be generated which I don't think is worth it.

And finally, we use **string resources** (so later it can be translated, if needed) for any text that is visible to the user and constants when dealing with the internal working of the application. For example, we have string constants for intent identifiers

devansh-299:
For the above reasons, I am closing this pull request, @Harshad09 if you have any doubts feel free to ask

Harshad09:
> For the above reasons, I am closing this pull request, @Harshad09 if you have any doubts feel free to ask


I actually took the reference from create Si activity where we have three inputs and we are showing particular toast. If you feel this is not needed then no worries @devansh-299 .

WhatsApp Image 2021-01-12 at 11 25 23 AM

PR #1340, opened 2022-08-07
Fixed Issue No #1198 Does not show client details.

### LINKED ISSUE #1198: Does not show client details
**Summary:**

The App not showing the client details

**Steps to reproduce:**

Just go to the MainActivity and see here the showclientdetails() function. There is the code that is commented out.

**Expected behavior:**

Show the client details

**Screenshots:**
image

Prashant830:
## Issue Fix
Fixes #1198

## Screenshots
<p>

</p>

## Description
 In mainActivity, mainPresenter, BaseHomeContract interface I am going to remove the `showclientdetails()`  function.

##

### DIFF
--- a/mifospay/src/main/java/org/mifos/mobilewallet/mifospay/home/BaseHomeContract.java
+++ b/mifospay/src/main/java/org/mifos/mobilewallet/mifospay/home/BaseHomeContract.java
@@ -17,8 +17,6 @@ import okhttp3.ResponseBody;
 public interface BaseHomeContract {
 
     interface BaseHomeView extends BaseView<BaseHomePresenter> {
-
-        void showClientDetails(Client client);
     }
 
     interface BaseHomePresenter extends BasePresenter {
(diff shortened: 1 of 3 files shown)


mifospay/src/main/java/org/mifos/mobilewallet/mifospay/home/BaseHomeContract.java
```diff
@@ -17,8 +17,6 @@
 public interface BaseHomeContract {
 
     interface BaseHomeView extends BaseView<BaseHomePresenter> {
-
-        void showClientDetails(Client client);
```
devansh-299, line 21: Since we don't have any methods inside the interface, we can skip it altogether. If any requirement arises for the same in the future, we can always create it again

devansh-299 (changes requested):

mifospay/src/main/java/org/mifos/mobilewallet/mifospay/home/BaseHomeContract.java
Prashant830: @devansh-299  if we removed `BaseHomeView` Interface then we also required to removes the  `BaseHomePresenter`  Interface because  `BaseHomeView` Interface extend the property of `BaseHomePresenter`.

mifospay/src/main/java/org/mifos/mobilewallet/mifospay/home/BaseHomeContract.java
```diff
@@ -17,8 +17,6 @@
...
     }
 
     interface BaseHomePresenter extends BasePresenter {
```
Prashant830, line 22: @devansh-299 by chance if we removed both of the interfaces (`BaseHomeView` , `BaseHomePresenter`).
then where we add `void fetchClientDetails();` method??

devansh-299 (approved):

PR #1374, opened 2023-03-27
Restructured input fields in Edit Profile Screen

### LINKED ISSUE #1334: Wrong text field labels on edit profile screen
**Summary:**
Instead of showing actual labels, the previous values of fields are used as labels.

**Screenshots:**
image

vishal203202:
## Issue Fix
Fixes #1334

## Screenshots
Screenshot_20230327_233804_Mifos Pay

## Description

Initially, the input fields were set up to display the user's credentials as hints. This was intended to show the user their details when they opened their edit profile screen for the first time. However, as the hints are also associated with the floating labels, the floating text fields were showing the wrong information. The solution was to implement a simple TextView with an EditText to serve the same purpose.

##

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
-FROM openjdk:8-jdk
+FROM openjdk:11-jdk
 
 ENV GRADLE_HOME /opt/gradle
-ENV GRADLE_VERSION 4.6
+ENV GRADLE_VERSION 6.7.1
 
 RUN set -o errexit -o nounset \
 	&& echo "Downloading Gradle" \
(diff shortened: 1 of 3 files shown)


rchtgpt (changes requested):
Hi @BelieveInTheLimitless, thanks for your contribution. This is a great start. There are a few things we can improve in this Pull Request:
* This should have been a point of discussion before you started working on this issue but the ideal behavior of the editText in this case should be to **not have any floating label at all**. Since we already have a helper text below the editText, we don't want a floating label to display the same thing. Can you please modify your PR in a way that completely removes the floating label?
* I see changes in files such as the `Dockerfile` where you have updated the `GRADLE_VERSION` and the `openjdk` version. Can you please avoid committing such changes that are not related to the solution at hand? You will probably have to revert this committed change or send a completely new PR.
* According to [Mifos PR guidelines](https://github.com/openMF/mobile-wallet/blob/master/.github/CONTRIBUTING.md#pull-request-guidelines), each PR must only be one final commit. It'd be great if you can squash your commits at the end.
* Please include a screen recording of your solution instead of just a screenshot. That would make it easier for us to visualize your changes.

PR #1378, opened 2023-06-06
Fix: Add trim functionality to EditTexts

### LINKED ISSUE #423: Add trim() method to all EditText objects in the project

rchtgpt:
## Issue Fix
Fixes #423 

## Description

Found and added the `trim()` function to all instances of EditTexts with the edge case of breaking if the user inputs a String with leading or trailing spaces. The only exceptions to this were password-related EditTexts & already-defined TextViews because I don't believe we should manipulate the user's inputted password.

##

### DIFF
--- a/mifospay/src/main/java/org/mifos/mobilewallet/mifospay/auth/ui/LoginActivity.java
+++ b/mifospay/src/main/java/org/mifos/mobilewallet/mifospay/auth/ui/LoginActivity.java
@@ -94,8 +94,8 @@ public class LoginActivity extends BaseActivity implements AuthContract.LoginVie
     }
 
     private void handleLoginInputChanged() {
-        String usernameContent = etUsername.getText().toString();
-        String passwordContent = etPassword.getText().toString();
+        String usernameContent = etUsername.getText().toString().trim();
+        String passwordContent = etPassword.getText().toString().trim();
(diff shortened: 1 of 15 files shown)


luckyman20:
The thought that we should not manipulate the user's password is correct, but we should warn the user not to include any spaces in the password. Can you please open an issue for that and send a PR as well?

luckyman20 (approved):
LGTM!

PR #1655, opened 2024-06-14
fix : moved notification to feature module

AdityaKumdale:
## Issue Fix
Fixes #{Issue Number}
Jira Task: [Task_Number](Task_Link)

##


feature/notification/build.gradle.kts
```diff
@@ -0,0 +1,16 @@
...
+android {
+    namespace = "org.mifospay.notification"
+}
+
+dependencies {
+    implementation(projects.core.data)
+    implementation(libs.androidx.appcompat)
+
+    //we need this for pull to refresh implementation
+    implementation("androidx.compose.material:material:1.6.0")
```
therajanmaurya: we should have this library in designsystem or UI module check and confirm

therajanmaurya: and never use hard coded library, always use from libs

therajanmaurya (changes requested):

feature/notification/build.gradle.kts
AdityaKumdale: Yes I will use libs now

PR #1679, opened 2024-06-27
refactor: new si to compose

PratyushSingh07:
Jira Ticket: https://mifosforge.jira.com/browse/MW-65


PratyushSingh07:
please review @AdityaKumdale @Aditya-gupta99 @AvneetSingh2001

mifospay/src/main/java/org/mifospay/standinginstruction/ui/NewSIScreen.kt
```diff
@@ -0,0 +1,410 @@
...
+                    .padding(horizontal = 20.dp),
+                value = vpa,
+                onValueChange = {
+                    vpa = it
+                },
+                label = R.string.vpa,
+                trailingIcon = {
+                    IconButton(onClick = { startScan() }) {
+                        Icon(
+                            imageVector = Icons.Filled.QrCode2,
```
AdityaKumdale: MifosIcons

mifospay/src/main/java/org/mifospay/standinginstruction/ui/StandingInstructionScreen.kt
```diff
@@ -42,67 +46,72 @@ fun StandingInstructionsScreen(
...
-                            amount = items.amount.toString(),
-                        )
+
+            is StandingInstructionsUiState.Error -> {
+                EmptyContentScreen(
+                    modifier = Modifier,
+                    title = stringResource(id = R.string.error_oops),
+                    subTitle = stringResource(id = R.string.error_fetching_si_list),
+                    iconTint = Color.Black,
+                    iconImageVector = Icons.Rounded.Info
```
AdityaKumdale: MifosIcons

mifospay/src/main/java/org/mifospay/standinginstruction/ui/StandingInstructionScreen.kt
```diff
@@ -42,67 +46,72 @@ fun StandingInstructionsScreen(
...
+                            )
+                        }
                     }
                 }
             }
-        }
 
+        }
     }
 }
```
AdityaKumdale, line 117: Can use previewparameter here

AdityaKumdale (changes requested):

mifospay/src/main/java/org/mifospay/standinginstruction/ui/NewSIScreen.kt
```diff
@@ -0,0 +1,410 @@
...
+                        stringResource(R.string.standing_instruction_created_successfully),
+                        Toast.LENGTH_SHORT
+                    ).show()
+                }
+            }
+        })
+}
+
+@Composable
+fun NewSIContent(
```
therajanmaurya: Why naming are same for above and below function? rename it with `NewSIBody`

mifospay/src/main/java/org/mifospay/standinginstruction/ui/NewSIScreen.kt
```diff
@@ -0,0 +1,410 @@
...
+    setCancelClicked: (Boolean) -> Unit,
+    confirm: (Long, Double, Int, String) -> Unit
+) {
+    val context = LocalContext.current
+
+    MifosScaffold(
+        topBarTitle = R.string.tile_si_activity,
+        backPress = { onBackPress.invoke() },
+        scaffoldContent = {
+            when (uiState) {
```
therajanmaurya: This flow seems wrong because always content form is visible by default and it will always be visible no matter what user action did like 
- Initially content form will be visible 
- If user click add or create etc then a progress bar will be visible on top of form but form will not go away but in below logic it will.
- We handle UI state on user action like pass, fail, validate state and show the toast etc.

therajanmaurya (changes requested):

PR #1693, opened 2024-07-03
Refactor : Use MaterialTheme colors for a uniform theme. This makes it easier to change themes.

Shermine237:
## Issue Fix
Fixes #{Issue Number}
Jira Task: [Task_Number](Task_Link)

##


Shermine237:
## If we want to change de default black color in MaterialTheme it's now easy. Just edit "mifospay/src/main/java/org/mifospay/theme/Theme.kt" and "core/designsystem/src/main/kotlin/org/mifospay/core/designsystem/theme/Theme.kt".
Then the theme color change will be applied to the entire application. This makes it easy to apply graphic charters.

## For green color, for example

PR #1755, opened 2024-09-01
feat: Ktor implementation for auth service

PratyushSingh07:


core/network/build.gradle.kts
```diff
@@ -61,6 +61,8 @@ dependencies {
     implementation("io.ktor:ktor-client-json:2.3.4")
     implementation("io.ktor:ktor-client-websockets:2.3.4")
     implementation("io.ktor:ktor-serialization-kotlinx-json:2.3.4")
+    implementation("org.jetbrains.kotlinx:kotlinx-serialization-json:1.3.0")
```
therajanmaurya: We have serialization plugin in model module and model module should only contains network models. 

Why did you add the serialization in network module?

core/network/src/main/kotlin/org/mifospay/core/network/di/NetworkModule.kt
```diff
@@ -177,8 +177,8 @@ class NetworkModule {
 
     @Provides
     @Singleton
-    fun provideAuthenticationService(client: HttpClient): KtorAuthenticationService {
-        return KtorAuthenticationService(client)
+    fun provideAuthenticationService(client: HttpClient, preferencesHelper: PreferencesHelper): KtorAuthenticationService {
```
therajanmaurya: Don't we have the datastore? @AdityaKumdale recently migrated the SharedPrefrence with Datastore.

therajanmaurya (changes requested):

core/network/build.gradle.kts
PratyushSingh07: this is done

core/network/src/main/kotlin/org/mifospay/core/network/di/NetworkModule.kt
PratyushSingh07: will have to crosscheck this with aditya

AdityaKumdale: Yes we have implemented proto datastore instead of preference store

therajanmaurya: I don't see any implementation of datastore in `datastore` module @AdityaKumdale why?

PR #1870, opened 2025-05-23
Fix Client and Merchant registration error

Shreyash16b:
## Issue Fix
Fixes MW-219 Issue on client and merchant registration
Jira Task: [219](https://mifosforge.jira.com/browse/MW-219)

## Screenshots

| Duplicate Mobile No. | Weak Password | Password with space/repititions
|----------------------|---------------|---------------|
|  |  |  |

### Mobile No. 
- HTTP response BODY contained 
    ```{"developerMessage":"The request caused a data integrity issue to be fired by the database.","httpStatusCode":"403","defaultUserMessage":"Client with mobileNo `1234567890` already exists","userMessageGlobalisationCode":"error.msg.client.duplicate.mobileNo","errors":[{"defaultUserMessage":"Client with mobileNo `1234567890` already exists","parameterName":"mobileNo","developerMessage":"Client with mobileNo `1234567890` already exists","userMessageGlobalisationCode":"error.msg.client.duplicate.mobileNo","args":[{"value":"1234567890"}]}]}```

- Which means the database expects unique Mobile No.

- Non-unique caused KtorClient error 
```Illegal input: Fields [officeId, clientId, resourceId, savingsId] are required for type with serial name 'org.mifospay.core.network.model.ClientResponseEntity', but they were missing at path: $```

- So, added code to ensure unique mobile number along with username in function ```checkUniqueFields```

### Password
- HTTP response BODY contained
```{"developerMessage":"The request was invalid. This typically will happen due to validation errors which are provided.","httpStatusCode":"400","defaultUserMessage":"Validation errors exist.","userMessageGlobalisationCode":"validation.msg.validation.errors.exist","errors":[{"defaultUserMessage":"Password must be 12 to 50 characters long, containing at least one uppercase letter, one lowercase letter, one numeric digit, and one special character, with no spaces or consecutive repeating characters","parameterName":"password","developerMessage":"Password must be 12 to 50 characters long, containing at least one uppercase letter, one lowercase letter, one numeric digit, and one special character, with no spaces or consecutive repeating characters","userMessageGlobalisationCode":"validation.msg.user.password.does.not.match.regexp","args":[{"value":"Abcdef123456#@@"},{"value":"^(?!.*(.)\\1)(?!.*\\s)(?=.*\\d)(?=.*[a-z])(?=.*[A-Z])(?=.*[^\\w\\s]).{12,50}$"}]}]}```

- Which means the database expects the password to follow these requirements : 
  - Must be between 12 and 50 characters long (not checked properly)
  - At least 1 uppercase character
  - At least 1 lowercase character
  - At least 1 numeric digit
  - At least 1 special character
  - No spaces (not checked for)
  - No consecutive repeating characters (not checked for)

- Weak password caused KtorClient error
```Illegal input: Field 'resourceId' is required for type with serial name 'org.mifospay.core.network.model.CommonResponse', but it was missing at path: $```

- Modified the ```PasswordChecker.kt``` to consider all requirements
- Modified the DialogBox error in ```SignUpViewModel.kt``` to display the requirements if password is weak
##

- [✅] Run the unit tests with `./gradlew check` to make sure you didn't break anything

- [✅ ] If you have multiple commits please combine them into one commit by squashing them.

### DIFF
--- a/core/ui/src/commonMain/kotlin/org/mifospay/core/ui/utils/PasswordChecker.kt
+++ b/core/ui/src/commonMain/kotlin/org/mifospay/core/ui/utils/PasswordChecker.kt
@@ -9,6 +9,8 @@
  */
 package org.mifospay.core.ui.utils
 
+import org.mifospay.core.common.utils.hasConsecutiveRepetitions
+import org.mifospay.core.common.utils.hasSpaces
 import kotlin.math.log2
 import kotlin.math.pow
 
@@ -19,27 +21,20 @@ object PasswordChecker {
     private const val MAX_PASSWORD_LENGTH = 50
 
     fun getPasswordStrengthResult(password: String): PasswordStrengthResult {
-        when {
-            password.isEmpty() -> return PasswordStrengthResult.Error("Password cannot be empty.")
-            password.length > MAX_PASSWORD_LENGTH -> {
-                return PasswordStrengthResult.Error(
-                    "Password is too long. Maximum length is $MAX_PASSWORD_LENGTH characters.",
-                )
-            }
-            hasSpaceOrConsecutiveRepetitions(password) -> {
-                return PasswordStrengthResult.Error(
-                    "Password must not contain spaces or repeating characters.",
-                )
-            }
+        val errors = buildList {
+            if (password.isEmpty()) add("- Password cannot be empty.")
+            if (password.length > MAX_PASSWORD_LENGTH) add("- Password is too long. Maximum length is $MAX_PASSWORD_LENGTH characters.")
+            if (password.hasSpaces()) add("- Password must not contain spaces.")
+            if (password.hasConsecutiveRepetitions()) add("- Password must not contain consecutive repetitive characters.")
+        }
+
+        if (errors.isNotEmpty()) {
+            return PasswordStrengthResult.Error(errors.joinToString("\n"))
         }
 
         val result = getPasswordStrength(password)
 
         return PasswordStrengthResult.Success(result)
-    }
-
-    fun hasSpaceOrConsecutiveRepetitions(password: String): Boolean {
-        return Regex("(.)\\1").containsMatchIn(password) || password.contains(" ")
     }
 
     private fun getPasswordStrength(password: String): PasswordStrength {
@@ -91,10 +86,10 @@ object PasswordChecker {
         if (password.length < STRONG_PASSWORD_LENGTH) {
             feedback.add("For a stronger password, use at least $STRONG_PASSWORD_LENGTH characters.")
         }
-        if (Regex("(.)\\1").containsMatchIn(password)) {
+        if (password.hasConsecutiveRepetitions()) {
             feedback.add("Remove consecutive repeating characters.")
         }
-        if (password.contains(" ")) {
+        if (password.hasSpaces()) {
             feedback.add("Remove spaces.")
         }
 
--- a/core/ui/src/commonMain/kotlin/org/mifospay/core/ui/utils/PasswordChecker.kt
+++ b/core/ui/src/commonMain/kotlin/org/mifospay/core/ui/utils/PasswordChecker.kt
@@ -22,10 +22,10 @@ object PasswordChecker {
 
     fun getPasswordStrengthResult(password: String): PasswordStrengthResult {
         val errors = buildList {
-            if (password.isEmpty()) add("- Password cannot be empty.")
-            if (password.length > MAX_PASSWORD_LENGTH) add("- Password is too long. Maximum length is $MAX_PASSWORD_LENGTH characters.")
-            if (password.hasSpaces()) add("- Password must not contain spaces.")
-            if (password.hasConsecutiveRepetitions()) add("- Password must not contain consecutive repetitive characters.")
+            if (password.isEmpty()) add("Password cannot be empty.")
+            if (password.length > MAX_PASSWORD_LENGTH) add("Password is too long. Maximum length is $MAX_PASSWORD_LENGTH characters.")
+            if (password.hasSpaces()) add("Password must not contain spaces.")
+            if (password.hasConsecutiveRepetitions()) add("Password must not contain consecutive repetitive characters.")
         }
 
         if (errors.isNotEmpty()) {
--- a/core/ui/src/commonMain/kotlin/org/mifospay/core/ui/utils/PasswordChecker.kt
+++ b/core/ui/src/commonMain/kotlin/org/mifospay/core/ui/utils/PasswordChecker.kt
@@ -13,10 +13,10 @@ import kotlin.math.log2
 import kotlin.math.pow
 
 object PasswordChecker {
-    private const val MIN_PASSWORD_LENGTH = 8
-    private const val STRONG_PASSWORD_LENGTH = 12
+    private const val MIN_PASSWORD_LENGTH = 12
+    private const val STRONG_PASSWORD_LENGTH = 16
     private const val MIN_ENTROPY_BITS = 60.0
-    private const val MAX_PASSWORD_LENGTH = 128
+    private const val MAX_PASSWORD_LENGTH = 50
 
     fun getPasswordStrengthResult(password: String): PasswordStrengthResult {
         when {
@@ -33,6 +33,10 @@ object PasswordChecker {
         return PasswordStrengthResult.Success(result)
     }
 
+    fun hasSpaceOrConsecutiveRepetitions(password: String): Boolean {
+        return Regex("(.)\\1").containsMatchIn(password) || password.contains(" ")
+    }
+
     fun getPasswordStrength(password: String): PasswordStrength {
         val length = password.length
         val hasUpperCase = password.any { it.isUpperCase() }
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: core/ui/src/commonMain/kotlin/org/mifospay/core/ui/utils/PasswordChecker.kt [hunk 1]
```diff
@@ -33,6 +33,10 @@ object PasswordChecker {
         return PasswordStrengthResult.Success(result)
     }
 
+    fun hasSpaceOrConsecutiveRepetitions(password: String): Boolean {
+        return Regex("(.)\\1").containsMatchIn(password) || password.contains(" ")
+    }
+
     fun getPasswordStrength(password: String): PasswordStrength {
```
HekmatullahAmin: if `getPasswordStrength` is used only in this file make it private

core/ui/src/commonMain/kotlin/org/mifospay/core/ui/utils/PasswordChecker.kt
```diff
@@ -47,12 +51,14 @@ object PasswordChecker {
         return when {
             length < MIN_PASSWORD_LENGTH -> PasswordStrength.LEVEL_0
             numTypesPresent == 1 -> PasswordStrength.LEVEL_1
-            numTypesPresent == 2 -> PasswordStrength.LEVEL_2
-            numTypesPresent == 3 && length >= STRONG_PASSWORD_LENGTH -> PasswordStrength.LEVEL_4
+            numTypesPresent == 2 || numTypesPresent == 3 ||
+                hasSpaceOrConsecutiveRepetitions(password) -> PasswordStrength.LEVEL_2
```
HekmatullahAmin: the space and the repetition is error and it should not make the decision that our password is weak or strong. so it can be handled at the `getPasswordStrengthResult` inside the when block you can create another that says `PasswordStrengthResult.Error` for space and consective repeating characters. 
This is a personal idea

core/ui/src/commonMain/kotlin/org/mifospay/core/ui/utils/PasswordChecker.kt
```diff
@@ -47,12 +51,14 @@ object PasswordChecker {
...
             numTypesPresent == 4 && length >= STRONG_PASSWORD_LENGTH &&
                 entropyBits >= MIN_ENTROPY_BITS -> PasswordStrength.LEVEL_5
+            numTypesPresent == 4 && length >= STRONG_PASSWORD_LENGTH -> PasswordStrength.LEVEL_4
+            numTypesPresent == 4 && length < STRONG_PASSWORD_LENGTH -> PasswordStrength.LEVEL_3
 
-            else -> PasswordStrength.LEVEL_3
+            else -> PasswordStrength.LEVEL_2
```
HekmatullahAmin: put the else block to `LEVEL3` because if:
- numTypesPresetn == 2 so the level will be LEVEL2 (no need for `hasSpace...`)
- numTypesPresent == 4 and length < Strong password then again will fall to else block

feature/auth/src/commonMain/kotlin/org/mifospay/feature/auth/signup/SignupViewModel.kt
```diff
@@ -347,50 +368,62 @@ class SignupViewModel(
...
+                "mobileNo" to "Mobile Number",
             )
 
-            when (result) {
-                is DataState.Error -> {
-                    val message = result.exception.message.toString()
-                    mutableStateFlow.update {
-                        it.copy(dialogState = SignUpDialog.Error(message))
+            for ((fieldName, fieldValue) in fields) {
+                val result = searchRepository.searchResources(
```
HekmatullahAmin: as what i read your saying we received the error cause the mobile is not unique. 
so for checking both `username` and `mobile` use can use `asyn{} and awaitAll{}` instead of using `for loop`. 
- use `async{}` to launch coroutine so that both run in parallel. and then use `awaitAll` to wait for both of them to complete.
- don't create `fieldsToCheck` and `fieldNamesMap` always when you see some extra code is added nearly achieving same look how you can make it minimum. 
- belwo is an example change it accordingly if needed:
```
private fun initiateSignUp() {
    showLoadingDialog()

    val fieldsToCheck = mapOf(
        "Username" to state.userNameInput,
        "Mobile Number" to state.mobileNumberInput
    )

    checkUniqueFields(fieldsToCheck)
}

private fun checkUniqueFields(fields: Map<String, String>) {
    viewModelScope.launch {
        val results = fields.map { (label, value) ->
            async {
                val result = searchRepository.searchResources(value, Constants.CLIENTS, false)
                label to result
            }
        }.awaitAll()

        val errorMessages = results.mapNotNull { (label, result) ->
            when (result) {
                is DataState.Success -> {
                    if (result.data.isNotEmpty()) "$label already exists." else null
                }
                is DataState.Error -> result.exception.localizedMessage
                    ?: "Error checking $label."
                else -> null
            }
        }

        if (errorMessages.isNotEmpty()) {
            showErrorDialog(errorMessages.joinToString("\n"))
        } else {
            proceedWithUserCreation()
        }
    }
}

```

Shreyash16b:
@HekmatullahAmin Please review the changes I have made

core/ui/src/commonMain/kotlin/org/mifospay/core/ui/utils/PasswordChecker.kt
```diff
@@ -26,14 +26,23 @@ object PasswordChecker {
                     "Password is too long. Maximum length is $MAX_PASSWORD_LENGTH characters.",
                 )
             }
+            hasSpaceOrConsecutiveRepetitions(password) -> {
```
HekmatullahAmin: having extension function here will be good here the code will be clean and also avoid the duplication of using `Regex("(.)\\1").containsMatchIn`

```
// Extension function to check spaces or consecutive repetitions
fun String.hasSpaceOrConsecutiveRepetitions(): Boolean {
    return this.contains(" ") || this.hasConsecutiveRepetitions()
}

fun String.hasConsecutiveRepetitions(): Boolean {
    return Regex("(.)\\1").containsMatchIn(this)
}
```

- and then you can call it like this: which will look the same as the other two at the top
```
password.hasSpaceOrConsecutiveRepetitions() -> {
                return PasswordStrengthResult.Error(
                    "Password must not contain spaces or repeating characters.",
                )
            }
```

core/ui/src/commonMain/kotlin/org/mifospay/core/ui/utils/PasswordChecker.kt
```diff
@@ -82,6 +91,12 @@ object PasswordChecker {
         if (password.length < STRONG_PASSWORD_LENGTH) {
             feedback.add("For a stronger password, use at least $STRONG_PASSWORD_LENGTH characters.")
         }
+        if (Regex("(.)\\1").containsMatchIn(password)) {
```
HekmatullahAmin: - In my another review i told you if you see any duplicate code see if you can with an approach that will be simple and yet look clean this is one case of it. with doing this we will avoid duplication and our code looks clean and we can also unit test it
```
if (password.hasConsecutiveRepetitions()) {
            feedback.add("Remove consecutive repeating characters.")
        }
```

core/ui/src/commonMain/kotlin/org/mifospay/core/ui/utils/PasswordChecker.kt
```diff
@@ -47,10 +56,10 @@ object PasswordChecker {
         return when {
             length < MIN_PASSWORD_LENGTH -> PasswordStrength.LEVEL_0
             numTypesPresent == 1 -> PasswordStrength.LEVEL_1
-            numTypesPresent == 2 -> PasswordStrength.LEVEL_2
-            numTypesPresent == 3 && length >= STRONG_PASSWORD_LENGTH -> PasswordStrength.LEVEL_4
+            numTypesPresent == 2 || numTypesPresent == 3 -> PasswordStrength.LEVEL_2
```
HekmatullahAmin, line 54: Let's keep it like it is coded like before if `numTypesPresent == 3` then it will fall to else block which will be `PasswordStrength.LEVEL_3` unless you have a reason that you did it that way?

HekmatullahAmin:
@Shreyash16b good job. just fix suggestions. And also in your third screenshort `Password with space/repititions` your password is `abcdef123456` which does not have any space nor repitition how this error dialog accured?

core/ui/src/commonMain/kotlin/org/mifospay/core/ui/utils/PasswordChecker.kt
Shreyash16b: That causes error because if numTypesPresent ==3 falls into the else block, then it is assigned PasswordStrength.LEVEL_3 which corresponds to a PasswordStrengthState.GOOD (inside the SignUpState.kt) which accepts the user input(doesn't show weak password and allows user to continue), but the backend responds with an error.
And about the spaces/repititions screenshot, it is because the space is in the end of the password. I'll update that screenshot.

HekmatullahAmin: okay good

Shreyash16b:
@HekmatullahAmin I have made the changes.
1. Used String Extension functions for checking spaces/repititions.
2. And now, I thought showing only relevant passwordError might be better for UX, So, it only shows the error which is present in the entered password instead of ```Password must not contains spaces or repititions```

HekmatullahAmin:
@niyajali if you have any suggestions let @Shreyash16b know otherswise kindly approve and then merge

niyajali (approved):

PR #1871, opened 2025-06-03
test: Add Unit Test for `:feature:auth` module

Shreyash16b:
## Issue Fix
Fixes 221
Jira Task: [221](https://mifosforge.jira.com/browse/MW-221)

## Screenshots
LoginViewModelTest Result
SignUpViewModelTest Result
MobileVerificationViewModelTest Result

## Description
1. Added dependency ([mokkery](mokkery.dev)) for unit tests.
2. Wrote tests for LoginViewModel.kt, SignUpViewModel.kt, MobileVerificationViewModel.kt.
3. These tests run on the Android and Desktop platforms, but give errors for JS and WASM.

This is the error while running command ``` ./gradlew :feature:auth:jsTest ```
JS Error

This is the error while running command ``` ./gradlew :feature:auth:wasmJsBrowserTest ```
WASM Error

##

- [✅] Run the unit tests with `./gradlew check` to make sure you didn't break anything

- [✅] If you have multiple commits please combine them into one commit by squashing them.


feature/auth/src/commonMain/kotlin/org/mifospay/feature/auth/signup/SignupViewModel.kt
```diff
@@ -239,57 +258,57 @@ class SignupViewModel(
     private fun handleSubmitClick() = when {
         state.savingsProductId == 0 -> {
             mutableStateFlow.update {
-                it.copy(dialogState = SignUpDialog.Error("Please select a savings account."))
+                it.copy(dialogState = SignUpDialog.Error(SELECT_SAVINGS_ACCOUNT))
```
HekmatullahAmin: Revert it back to previous as we don't use `Constants` for these messages we are gonna use `stringResource` which if we target different audience we will use localization

feature/auth/build.gradle.kts
```diff
@@ -43,5 +56,17 @@ kotlin {
 
             implementation(libs.play.services.auth)
         }
+
+        commonTest.dependencies {
+            implementation(kotlin("test-common"))
+            implementation(kotlin("test-annotations-common"))
+            implementation(libs.kotlinx.coroutines.test)
```
HekmatullahAmin: - Add the dependencies if it is needs for example `implementation(libs.kotlinx.coroutines.test)` this one is already added in `KMPLibraryConventionPlugin.kt` or `implementation(compose.uiTest)` as we are not focusing in UI test don't add it and other dependecnies that you added.
- @niyajali should he use `mokerry` third party library for testing?

feature/auth/src/commonTest/kotlin/org/mifospay/feature/auth/login/LoginViewModelTest.kt
```diff
@@ -0,0 +1,185 @@
...
+        return runTest(testDispatcher) {
+            assertEquals("", viewModel.stateFlow.value.username)
+            assertEquals("", viewModel.stateFlow.value.password)
+            assertFalse(viewModel.stateFlow.value.isPasswordVisible)
+            assertEquals(null, viewModel.stateFlow.value.dialogState)
+        }
+    }
+
+    @Test
+    fun `username change updates state`() = runTest(testDispatcher) {
```
HekmatullahAmin: use one style for naming the test. 
for example `loginViewModel_UsernameChanged_UpdatesState`

feature/auth/src/commonTest/kotlin/org/mifospay/feature/auth/login/LoginViewModelTest.kt
```diff
@@ -0,0 +1,185 @@
...
+            clientRepository = mockClientRepository,
+            userPreferencesRepository = mockUserPreferencesRepository,
+        )
+        viewModel = LoginViewModel(loginUseCase, SavedStateHandle())
+
+        viewModel.trySendAction(LoginAction.UsernameChanged("testuser"))
+        viewModel.trySendAction(LoginAction.PasswordChanged("testpass"))
+        viewModel.trySendAction(LoginAction.LoginClicked)
+
+        println(viewModel.stateFlow.value)
```
HekmatullahAmin: remove the `printlns`

core/data/src/commonMain/kotlin/org/mifospay/core/data/util/Constants.kt
```diff
@@ -52,4 +52,33 @@ object Constants {
     const val FETCH_CLIENT_DETAILS_USE_CASE = "Fetch Client Details UseCase"
     const val ERROR_FETCHING_NOTIFICATIONS = "Error fetching notifications"
     const val UNAUTHORIZED_ERROR = "401 Unauthorized"
+
+    const val SELECT_SAVINGS_ACCOUNT = "Please select a savings account."
+    const val ENTER_FIRST_NAME = "Please enter your first name."
```
HekmatullahAmin: remove these added these constants as i mentioned the reason below

HekmatullahAmin:
@niyajali if you have any suggestions for him regarding to which format or how to write unit testing regarding our format let him know

feature/auth/build.gradle.kts
```diff
@@ -11,6 +23,7 @@ plugins {
     alias(libs.plugins.mifospay.cmp.feature)
     alias(libs.plugins.kotlin.parcelize)
     alias(libs.plugins.kotlin.serialization)
+    id("dev.mokkery") version "2.7.2"
```
biplab1: This direct version dependency should be replaced. Please use `alias(libs.mokkery)` instead to keep it consistent with the version catalog.

feature/auth/src/commonMain/kotlin/org/mifospay/feature/auth/signup/SignupViewModel.kt
```diff
@@ -275,7 +275,7 @@ class SignupViewModel(
 
         state.mobileNumberInput.isEmpty() -> {
             mutableStateFlow.update {
-                it.copy(dialogState = SignUpDialog.Error("Please enter your mobile number."))
+                it.copy(dialogState = SignUpDialog.Error("Please enter a your mobile number."))
```
HekmatullahAmin: adding `a` is a grammatical mistake this should not ne added

feature/auth/src/commonMain/kotlin/org/mifospay/feature/auth/signup/SignupViewModel.kt
```diff
@@ -313,11 +313,13 @@ class SignupViewModel(
 
         !state.isPasswordStrong -> {
             val errorMessage = state.passwordError?.takeIf { it.isNotBlank() }
-                ?: "Please ensure password contains :" +
-                "\n- At least one uppercase character" +
-                "\n- At least one lowercase character" +
-                "\n- At least one numeric digit" +
-                "\n- At least one special character"
+                ?: """
```
HekmatullahAmin: don't remove the `\n` otherwise it will show all message attach to each other

revert any changes regarding the text

feature/auth/src/commonMain/kotlin/org/mifospay/feature/auth/signup/SignupViewModel.kt
```diff
@@ -349,13 +351,13 @@ class SignupViewModel(
 
         state.countryInput.isEmpty() -> {
             mutableStateFlow.update {
-                it.copy(dialogState = SignUpDialog.Error("Please enter your country."))
+                it.copy(dialogState = SignUpDialog.Error("Please enter your country"))
```
HekmatullahAmin: just one thing keep the text as the way were before

HekmatullahAmin:
@Shreyash16b fix the things that i mentioned and give me the authority to push changes to this PR. once i get the change i will pull it

HekmatullahAmin:
@Shreyash16b don't push any new code for now. just give me the authority so i do it

Shreyash16b:
@HekmatullahAmin Sorry,  I just pushed the code a minute before your message.

HekmatullahAmin:
- @Shreyash16b i can't change the PR title. could u please change it to this: `Add feature/auth unit tests and add KDoc for clarity`
- And also add these photos in the description of PR remove those error images that u received and posted.

# Description
This PR improves test coverage and developer experience in the authentication feature by:

- ✅ Added unit tests for:

    - SignUpViewModelTest

    - LoginViewModelTest

    - MobileVerificationViewModelTest

- 🔧 Mocking approach:

    - Used Mokkery to mock repositories in SignUpViewModelTest and MobileVerificationViewModelTest.

    - Mocked LoginUseCase in LoginViewModelTest for isolated and deterministic behavior.

- 🛠️ Test support enhancements:

    - Introduced the @OpenForMokkery annotation and applied the kotlin-allopen plugin scoped to this module, allowing open classes for mocking without polluting production codebase.

    - Created and integrated a StringProvider interface to decouple direct usage of getString() in ViewModels.

        - This enables test-safe injection of fake/mock StringProvider implementations.

        - In production, DefaultStringProvider uses the real Compose getString() call.

- 📚 Documentation:

    - Added comprehensive KDoc comments to improve test and utility class readability and maintainability.

## Android
## SignUpViewModelTest
image

## LoginViewModelTest
image

## MobileVerificationViewModelTest
image

## Ios
ios


## Desktop
desktop

## WasmJsBrowser
wasmJsBrowser

## JsNode
JsNode

HekmatullahAmin (approved):

feature/auth/src/commonTest/kotlin/org/mifospay/feature/auth/fakes/FakeSearchResult.kt
```diff
@@ -0,0 +1,21 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ *
+ * See https://github.com/openMF/mobile-wallet/blob/master/LICENSE.md
+ */
+package org.mifospay.feature.auth.fakes
+
+import org.mifospay.core.model.search.SearchResult
+
+val fakeSearchResult = SearchResult(
```
niyajali: Move this directly on that file where it has been used no need to create separate file for this

feature/auth/src/commonMain/kotlin/org/mifospay/feature/auth/signup/SignupViewModel.kt
```diff
@@ -346,7 +425,9 @@ class SignupViewModel(
         Enhancement: Move the following code in to a Use Case
      */
     private fun initiateSignUp() {
-        DialogManager.showLoading()
```
niyajali, line 349: When we introduced DialogManager, @HekmatullahAmin are you sure that these changes are necessary

feature/auth/src/commonTest/kotlin/org/mifospay/feature/auth/fakes/FakeUserInfo.kt
```diff
@@ -0,0 +1,26 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ *
+ * See https://github.com/openMF/mobile-wallet/blob/master/LICENSE.md
+ */
+package org.mifospay.feature.auth.fakes
+
+import org.mifospay.core.model.user.UserInfo
+
+val fakeUserInfo = UserInfo(
```
niyajali: this too and make it internal

feature/auth/src/commonTest/kotlin/org/mifospay/feature/auth/login/LoginViewModelTest.kt
```diff
@@ -0,0 +1,179 @@
...
+            savedStateHandle = SavedStateHandle(),
+        )
+    }
+
+    @AfterTest
+    fun tearDown() {
+        Dispatchers.resetMain()
+    }
+
+    // --------------------------------------------------------------------------
```
niyajali: remove this comments

feature/auth/src/commonTest/kotlin/org/mifospay/feature/auth/login/LoginViewModelTest.kt
```diff
@@ -0,0 +1,179 @@
...
+    // Boundary Case
+    // --------------------------------------------------------------------------
+
+    /**
+     * Verifies initial state values are set correctly on ViewModel init.
+     */
+    @Test
+    fun loginViewModel_InitialState_ValidInitialConditions() = runTest(testDispatcher) {
```
niyajali: remove the prefix from all test methods, and try to followT, est method should follow the `given_when_then` or `when_then` format.

feature/auth/src/commonTest/kotlin/org/mifospay/feature/auth/login/LoginViewModelTest.kt
```diff
@@ -0,0 +1,179 @@
...
+
+    /**
+     * Tests the toggle password visibility logic.
+     */
+    @Test
+    fun loginViewModel_TogglePasswordVisibility_UpdatesVisibilityState() = runTest(testDispatcher) {
+        assertFalse(viewModel.stateFlow.value.isPasswordVisible)
+
+        viewModel.trySendAction(LoginAction.TogglePasswordVisibility)
+        advanceUntilIdle()
```
niyajali, line 102: we won't need this always, remove the `testDispatcher` from this lamda `runTest(testDispatcher)`  as you've defined in `setup` method, use `turbine` libray to test the `flow`

feature/auth/src/commonTest/kotlin/org/mifospay/feature/auth/mobileVerify/MobileVerificationViewModelTest.kt
```diff
@@ -0,0 +1,166 @@
...
+import org.mifospay.core.data.repository.SearchRepository
+import org.mifospay.feature.auth.fakes.fakeSearchResult
+import kotlin.test.AfterTest
+import kotlin.test.BeforeTest
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+
+@OptIn(ExperimentalCoroutinesApi::class)
+class MobileVerificationViewModelTest {
```
niyajali, line 35: Similarly apply those changes on these test classes

niyajali (changes requested):

feature/auth/src/commonTest/kotlin/org/mifospay/feature/auth/login/LoginViewModelTest.kt
niyajali: For reference - https://github.com/openMF/kmp-project-template/blob/dev/core-base/datastore/src/commonTest/kotlin/template/core/base/datastore/repository/DefaultReactiveUserPreferencesRepositoryTest.kt

niyajali: And this is not a KMP project, but you can check it out just for reference - https://github.com/devikontech/restaurant-billing-pos/tree/main/feature/addonitem/src/test/kotlin/com/niyaj/addonitem

feature/auth/src/commonMain/kotlin/org/mifospay/feature/auth/signup/SignupViewModel.kt
HekmatullahAmin: @niyajali Previously i created so to have one object handle the dialog logic in all our screens and cause of Parcelization we couldn't persist those of `StringResource` type. Although we can achieve the unit testing with DialogManager, but it will bound to lifecycle of viewmodel and also in unit testing then we also need to call it to check it is state.

feature/auth/src/commonTest/kotlin/org/mifospay/feature/auth/login/LoginViewModelTest.kt
HekmatullahAmin: i removed it, but we don't have any flow in the view models to test

niyajali: viewModel.stateFlow, viewModel.eventFlow are flowable property

HekmatullahAmin: Hi @niyajali 

Thanks for pointing that out! When you mentioned `flow` earlier, I initially thought of cold flows (like the ones from repositories or data sources), so I misunderstood your reference to `viewModel.stateFlow` and `eventFlow`. You're absolutely right — they are StateFlow and Flow, and can definitely be tested reactively using Turbine.

I checked the [Android Developers documentation on testing flows](https://developer.android.com/kotlin/flow/test#stateflows), and they actually recommend treating StateFlow as a data holder and asserting on its .value directly — especially when the logic is simple and we only care about the current state, not intermediate emissions (since StateFlow is conflated anyway).

That said, I'm totally okay with swapping to Turbine for consistency and better observation of flow emissions — especially for testing eventFlow, which is a one-shot flow of events.

So yeah, I value your thoughts on this — and it's clear that Turbine adds power when needed. I just wanted to share my thinking on why .value seemed sufficient so far.

Let me know if you'd like me to go ahead and convert the tests!
Screenshot 2025-07-13 at 11 26 11

niyajali: I understand your point, but testing can be inconsistent and really sucks sometimes in my experience, sometimes the UI displays correct output while tests fail, or tests pass locally in your IDE but fail in CI.

 This is why I'd suggest combining both approaches. When collecting values with `.value`, you might need to use `waitUntilIdle()` or collect the flow properly, since flows only emit values when there's an active collector. However, this might not be necessary in some cases as this project designed UDF (Unidirectional Data Flow) pattern.

 Let's implement what works best for our use case, and for reference, check the Now in Android project or other similar examples given above.

https://github.com/android/nowinandroid/blob/main/feature/foryou/src/test/kotlin/com/google/samples/apps/nowinandroid/feature/foryou/ForYouViewModelTest.kt

HekmatullahAmin: @niyajali 

Our BaseViewModel handles every user action asynchronously via viewModelScope.launch, using advanceUntilIdle() is necessary to ensure state changes are fully applied before we assert on stateFlow.value.
```
viewModelScope.launch {
    internalActionChannel.consumeAsFlow().collect { action -> handleAction(action) }
}
```
This means:

Even if handleAction updates the state immediately, it's still queued on a coroutine and will run in a future dispatch cycle.

So when your test does:
```
viewModel.trySendAction(...)
assertEquals(expected, viewModel.stateFlow.value)
```
...it might assert too early, before handleAction actually runs.

That's why you use `advanceUntilIdle()` — it gives the coroutine scheduler time to process any enqueued tasks before you assert.

- .value makes perfect sense when we only care about the final, conflated state.

- That said, Turbine is a great fit when we need to capture multiple state emissions over time.

So going forward, I’ll continue using .value + advanceUntilIdle() for most simple stateFlow assertions, but I’ll switch to using Turbine for eventFlow testing to ensure we properly collect and verify emitted events.

below is an example for both approach which i think using turbine here is not necessary our logic is more simple in first one:

```
@Test
    fun givenFirstName_whenInputChanged_thenFirstNameIsUpdated() = runTest {
        val firstName = "John"
        viewModel.trySendAction(SignUpAction.FirstNameInputChange(firstName))
        advanceUntilIdle() 
        assertEquals(firstName, viewModel.stateFlow.value.firstNameInput)
    }

    @Test
    fun givenFirstNames_whenInputChanged_thenFirstNameIsUpdated() = runTest {
        viewModel.stateFlow.test {
            // Skip initial state
            skipItems(1)

            // Act: send username change
            val firstName = "John"
            viewModel.trySendAction(SignUpAction.FirstNameInputChange(firstName))

            // Assert: username should now be "alice"
            val state = awaitItem()
            assertEquals(firstName, state.firstNameInput)
        }
    }
```

niyajali: Yes, I meant the same, refactor all the test cases accordingly and use proper naming conventions like `given_when_then` or `when_then` and let me know.

HekmatullahAmin: @niyajali yeah already used [given_when_then](https://medium.com/turkcell/enhancing-your-tests-the-given-when-then-naming-technique-for-unit-testing-4e821704dd8a) naming convention.
and also added unit test for `eventFlow`

HekmatullahAmin: I’ve directly tested the `stateFlow` using `.value` along with `advanceUntilIdle()`, as done in the `Now in Android` project. Since our logic is relatively simple and we only care about the latest state, using `Turbine` would’ve added unnecessary checks like `skipItems()` or `awaitItem()` that don’t add value in this case.

niyajali:
@HekmatullahAmin Upload Test results for all platforms

HekmatullahAmin:
> @HekmatullahAmin Upload Test results for all platforms

@niyajali done. i run one viewmodel test in one platform if i did all it would take a lot of space. and also could u just cut it and paste it in PR description i don't have authority to do it.

niyajali (approved):

PR #1872, opened 2025-06-11
[MW-225] feat(auth): Implement full Sign In & Sign Up flow with validation

HekmatullahAmin:
Jira Task: [MW-225](https://mifosforge.jira.com/browse/MW-225)

## Screenshots

https://github.com/user-attachments/assets/29712344-cd42-4379-a8d6-6ad8341ef83e

### What’s Achieved in This PR
1. Implemented full Sign In and Sign Up flow

- Covers both happy and unhappy user journeys.

- Seamlessly transitions from input to account creation and navigation.

2. Added robust user input validations

- Ensures mandatory fields like name, email, and mobile number are filled.

- Email is checked for correct format.

- Password and confirm password must match.

3. Improved error handling

- All validation and API errors are now clearly presented to the user.

- Replaced generic or null messages with user-friendly alternatives.

- Handles unique field validation (username and mobile) with clear feedback.

4. Password requirement hinting

- A help card shows password rules inline before submission.

- Prevents trial-and-error by guiding users upfront.

5. Updated pin code validation

- Removed strict 6-digit requirement for pin codes.

- Allows shorter lengths to support different country formats (e.g., 4- or 5-digit codes).

6. Country and state dropdown integration

- Integrated [stefanbinder/countries-states](https://github.com/stefanbinder/countries-states) JSON for populating country-state mapping.

- Countries and their respective states are loaded from countries.json and rendered as dropdowns in the signup form.

##

[MW-225]: https://mifosforge.jira.com/browse/MW-225?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


HekmatullahAmin:
@niyajali could you please have a look here

niyajali (changes requested):
@HekmatullahAmin Fetch Country & State from server and show as drop-down if any API available otherwise create a list of country and state and show as drop-down.

And instead of dialog for password feedback show a card below of the password indicator which should contain list of password feedback.

niyajali:
And update the PR title like `[MW-225] - The PR Title`

[MW-225]: https://mifosforge.jira.com/browse/MW-225?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

feature/auth/src/commonMain/kotlin/org/mifospay/feature/auth/signup/SignupScreen.kt
```diff
@@ -95,6 +111,10 @@ internal fun SignupScreen(
         }
     }
 
+    LaunchedEffect(Unit) {
```
niyajali: remove this and call inside of `init` block on ViewModel

feature/auth/src/commonMain/kotlin/org/mifospay/feature/auth/signup/SignupScreen.kt
```diff
@@ -317,25 +358,65 @@ private fun SignupScreenContent(
...
+
+                // Country Dropdown
+                ExposedDropdownBox(
+                    expanded = isCountryDropdownExpanded,
                     label = stringResource(Res.string.feature_auth_country),
+                    value = selectedCountry,
                     onValueChange = {
                         onAction(SignUpAction.CountryInputChange(it))
+                        // Clear state when country changes
+                        onAction(SignUpAction.StateInputChange(""))
```
niyajali: You can update in ViewModel, no need to send multiple action

feature/auth/src/commonMain/kotlin/org/mifospay/feature/auth/signup/Country.kt
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * See https://github.com/openMF/mobile-wallet/blob/master/LICENSE.md
+ */
+package org.mifospay.feature.auth.signup
+
+import kotlinx.serialization.Serializable
+
+@Serializable
+internal data class Country(
+    val name: String,
```
niyajali: add `countryCode` parameter

feature/auth/src/commonMain/kotlin/org/mifospay/feature/auth/signup/SignupViewModel.kt
```diff
@@ -151,7 +153,11 @@ class SignupViewModel(
 
             is SignUpAction.CountryInputChange -> {
                 mutableStateFlow.update {
-                    it.copy(countryInput = action.country)
+                    it.copy(
+                        countryInput = action.country,
+                        // reset state when country changes
+                        stateInput = "",
```
niyajali, line 183: see here you reseting the state, so you no need to send multiple action from UI

feature/auth/src/commonMain/kotlin/org/mifospay/feature/auth/signup/SignupViewModel.kt
```diff
@@ -509,6 +504,33 @@ class SignupViewModel(
...
+    @OptIn(ExperimentalResourceApi::class)
+    private fun loadCountriesFromJson() {
+        Logger.d("Tag-viewmodel loadCountriesFromJson called")
+        viewModelScope.launch {
+            try {
+                val json = Json {
+                    // This will skip unexpected fields
+                    ignoreUnknownKeys = true
+                }
+                val bytes = Res.readBytes("files/countries.json")
```
niyajali: Move this file to the data module and create a interface like `AssetRepository or LocalAssetRepository` and create required methods to load this data so you can use it in multiple places.

feature/auth/src/commonMain/kotlin/org/mifospay/feature/auth/signup/SignupViewModel.kt
```diff
@@ -529,27 +551,9 @@ data class SignUpState(
...
-            PasswordStrengthState.STRONG,
-            PasswordStrengthState.VERY_STRONG,
-            -> true
-        }
-
-    @IgnoredOnParcel
-    val isPasswordMatch: Boolean
-        get() = passwordInput == confirmPasswordInput
-}
+    val passwordFeedback: List<String> = emptyList(),
```
niyajali: Use `ImmutableList<T>` instead of `List<T>`, which is not stable for better performance. Add the `kotlinx.collection.immutable` and `kotlinx.serialization.json` dependencies to the KMP convention plugin to make them accessible across all modules.

niyajali (changes requested):

HekmatullahAmin:
@niyajali can you have a look at here i created the interface if the want the data from json then they can create a function accordingly

core/designsystem/src/commonMain/kotlin/org/mifospay/core/designsystem/component/TextField.kt
```diff
@@ -74,9 +74,10 @@ fun MifosOutlinedTextField(
         value = value,
         onValueChange = onValueChange,
         label = label,
+        isError = isError && isFocused,
         readOnly = readOnly,
         supportingText = {
-            if (isError) {
+            if (isError && isFocused) {
```
niyajali: Why we're checking textfeld is focused or not, the user has to click on each text field to see the error

core/designsystem/src/commonMain/kotlin/org/mifospay/core/designsystem/component/TextField.kt
```diff
@@ -168,10 +169,10 @@ fun MifosTextField(
             }
         },
         placeholder = placeholder,
-        supportingText = errorText?.let {
-            {
+        supportingText = {
```
niyajali: do like this, so it doesn't take space for `supportingText` field, otherwise this(`if (isError && isFocused && errorText != null) ` )  may be true but the `supportingText ` composable will intitalize initially
```
errorText?.let {
    { }
 }

File: core/ui/src/commonMain/kotlin/org/mifospay/core/ui/MifosPasswordField.kt [hunk 9]
```diff
@@ -91,9 +97,9 @@ fun MifosPasswordField(
                 ),
             ) {
                 val imageVector = if (showPassword) {
-                    MifosIcons.OutlinedVisibilityOff
```
niyajali, line 94: `MifosIcons.OutlinedVisibilityOff` should be prefered when we're showing the password

feature/auth/src/commonMain/kotlin/org/mifospay/feature/auth/signup/SignupScreen.kt
```diff
@@ -231,12 +250,15 @@ private fun SignupScreenContent(
                 onValueChange = {
                     onAction(SignUpAction.MobileNumberInputChange(it))
                 },
+                errorMessage = stringResource(Res.string.feature_auth_error_mobile_empty),
```
niyajali: As we're showing the error messages in a dialog, adding this additional message is not necessary or required.

niyajali (changes requested):

HekmatullahAmin:
@niyajali just wanted to clarify the UX expectations here:

Would you prefer:

- Show red outline + support text even when the field hasn't been focused yet (i.e. as soon as the screen loads),
- Show only a `*required` message below each field (no red border or red support text styling),
- Or some other validation behavior?

image

feature/auth/src/commonMain/kotlin/org/mifospay/feature/auth/signup/SignupScreen.kt
HekmatullahAmin: yeah you are right the user will see those on dialog, but that is after clicking the submit. So before submitting i wanted to user know that putting it empty is not acceptable. wanted to providing real-time feedback

core/ui/src/commonMain/kotlin/org/mifospay/core/ui/MifosPasswordField.kt
HekmatullahAmin: I swap it because:
- when the eye is open -> we can't see the password
- when the eye was shut (close) -> we can see the password it was vice versa.

you can have a look at the video when the password is not visible the eye is open, and when the password is visible the eye is shot.
So i think when the eye is shut it means you can't see. (Samarth pointed this out and I think it should be like this, what do u think?)

core/designsystem/src/commonMain/kotlin/org/mifospay/core/designsystem/component/TextField.kt
HekmatullahAmin: ok so you mean use `errorText?.let` and then inside it do the checking.

core/designsystem/src/commonMain/kotlin/org/mifospay/core/designsystem/component/TextField.kt
HekmatullahAmin: yeah i put it like that because if we don't then when user opens sign up screen all the fields will be in red. But they eventually go to each field even if they press a dialog will pop up to enter it. I saw some mater design the red appear after clicking or typing in the `TextField` so cause of that i did it like this.

niyajali:
@HekmatullahAmin What type of validation we had earlier

HekmatullahAmin:
> @HekmatullahAmin What type of validation we had earlier

@niyajali previously it was only dialog if someone click submit it was showing the field must not be empty

niyajali:
@HekmatullahAmin Okay so currently you're passing some errorMessage on the text field initially, and what is that message

HekmatullahAmin:
> @HekmatullahAmin Okay so currently you're passing some errorMessage on the text field initially, and what is that message

yes the message are all like this field cannot be empty. 
for example `username is required `
`address line 1 connot be empty`
...

and will only show if they are empty and is focusable, so if the go to another field it shouldn't show. if not focusable and the user visit the sign up screen for the first time all of those fields is going to look red with their error message.

do you want me to keep it or revert it back to before or u want sth else?

niyajali:
Those error messages we're showing on the dialog right, so do not need it, revert to the original

HekmatullahAmin:
@niyajali I revert it and also removed the isError from composables in SignUp Screen. Previously it was passed from the SignUp screen checking if those were empty but in `MifosTextField` the isError was not passed. 
- So as previously there were no red error container nor error message i removed them

core/data/src/commonMain/kotlin/org/mifospay/core/data/repositoryImp/AssetRepositoryImpl.kt
```diff
@@ -0,0 +1,36 @@
+/*
+ * Copyright 2025 Mifos Initiative
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ *
+ * See https://github.com/openMF/mobile-wallet/blob/master/LICENSE.md
+ */
+package org.mifospay.core.data.repositoryImp
```
biplab1: There is a typo in repositoryImp, it should be repositoryImpl (also in directory name)

core/data/src/commonMain/kotlin/org/mifospay/core/data/repositoryImp/AssetRepositoryImpl.kt
```diff
@@ -0,0 +1,36 @@
...
+import org.mifospay.core.model.utils.Country
+
+class AssetRepositoryImpl : AssetRepository {
+    @OptIn(ExperimentalResourceApi::class)
+    override suspend fun getCountriesWithStates(): DataState<Map<String, List<String>>> {
+        return try {
+            val json = Json { ignoreUnknownKeys = true }
+            val bytes = Res.readBytes("files/countries.json")
+            val jsonString = bytes.decodeToString()
+            val parsed = json.decodeFromString<List<Country>>(jsonString)
```
biplab1: `parsed` could be renamed as `countries`

core/data/src/commonMain/kotlin/org/mifospay/core/data/repositoryImp/AssetRepositoryImpl.kt
```diff
@@ -0,0 +1,36 @@
...
+            DataState.Success(
+                parsed.associate { country ->
+                    country.name to country.states.map { it.name }.ifEmpty { listOf("N/A") }
```
biplab1: The repository should return the raw list (even if it’s empty) and let the UI decide how to present that to the user.

HekmatullahAmin:
@niyajali i revert those changes just one thing more how you want the behavious of `PasswordStrenghtIndicator`:
- to be shown always when the user navigate to `SignUpScreen`?
- or to be shown only if it is clicked in the `password` field and hide when user navigate to another field?

@biplab1 thanks for pointing i will fix the typos

niyajali:
As we've created the indicator card those are only visible when the focus state has been changed.

core/data/src/commonMain/kotlin/org/mifospay/core/data/repositoryImp/AssetRepositoryImpl.kt
HekmatullahAmin: Yes, it would be a good idea to handle empty state lists in the UI if we wanted to show a different UI or a specific message for the empty case. But for now, our logic is simple—we just want to display a list of states for the selected country. If a country doesn't have states, we show “N/A.” Handling this in the repository keeps our UI cleaner by avoiding conditional checks there.

HekmatullahAmin:
@niyajali added the neccessary changes if looks good approve and merge

HekmatullahAmin:
@biplab1 could u please resolve your requested changes and approve

feature/auth/src/commonMain/kotlin/org/mifospay/feature/auth/signup/SignupViewModel.kt
```diff
@@ -285,44 +295,38 @@ class SignupViewModel(
...
         }
 
-        state.passwordInput.length < MIN_PASSWORD_LENGTH -> {
+        state.passwordInput.isEmpty() -> {
             mutableStateFlow.update {
                 it.copy(
                     dialogState = SignUpDialog.Error(
-                        "Password must be at least $MIN_PASSWORD_LENGTH characters long.",
+                        message = "The password field cannot be empty.",
                     ),
```
revanthkumarJ: won't we are using stringResources instead of directly using strings

feature/auth/src/commonMain/kotlin/org/mifospay/feature/auth/signup/SignupScreen.kt
```diff
@@ -317,25 +318,51 @@ private fun SignupScreenContent(
...
-                MifosOutlinedTextField(
-                    value = state.countryInput,
+                var isCountryDropdownExpanded by remember { mutableStateOf(false) }
+                var isStateDropdownExpanded by remember { mutableStateOf(false) }
+
+                val selectedCountry = state.countryInput
+                val selectedState = state.stateInput
+                val statesForSelectedCountry =
+                    state.countriesWithStates[selectedCountry] ?: emptyList()
+
```
itsPronay: shouldn't we include them inside the `SignUpState` and access them from there!?!

core/model/src/commonMain/kotlin/org/mifospay/core/model/utils/Country.kt
```diff
@@ -0,0 +1,29 @@
...
+package org.mifospay.core.model.utils
+
+import kotlinx.serialization.Serializable
+
+@Serializable
+data class Country(
+    val code2: String,
+    val code3: String,
+    val name: String,
+    val capital: String,
```
revanthkumarJ, line 19: i think it's better to add default values as in api result if any arg not found it will throw an error but if we add default values it won't throw error.

revanthkumarJ (changes requested):

feature/auth/src/commonMain/kotlin/org/mifospay/feature/auth/signup/SignupScreen.kt
HekmatullahAmin: yes that make sense i will do it

feature/auth/src/commonMain/kotlin/org/mifospay/feature/auth/signup/SignupViewModel.kt
HekmatullahAmin: yeah it can be done i will do it just need to take care because then our error will who two types of message one `StringResource` and other `String/

core/model/src/commonMain/kotlin/org/mifospay/core/model/utils/Country.kt
HekmatullahAmin: the data is static and those field is always present so casue of that no need to defaults.

feature/auth/src/commonMain/kotlin/org/mifospay/feature/auth/signup/SignupScreen.kt
HekmatullahAmin: @itsPronay done

feature/auth/src/commonMain/kotlin/org/mifospay/feature/auth/signup/SignupViewModel.kt
HekmatullahAmin: @revanthkumarJ done created a utility functino to take care of it and the user can pass `string` or `stringResource` and the dialog will display the message

core/data/src/commonMain/kotlin/org/mifospay/core/data/repositoryImp/AssetRepositoryImpl.kt
HekmatullahAmin: @biplab1 i handled the empty case scenario in UI casue now we will use a `TextField` if no state is available so user enter their own, otherwise it will list states from our static data in dropdown

HekmatullahAmin:
@itsPronay , @revanthkumarJ , @biplab1 when u are available have a look and resolve your conversation and kindly approve

biplab1 (approved):

core/ui/src/commonMain/kotlin/org/mifospay/core/ui/PasswordStrengthIndicator.kt
```diff
@@ -21,26 +23,246 @@ import androidx.compose.foundation.layout.Spacer
...
+                                    .background(
+                                        color = animatedIndicatorColor,
+                                        shape = RoundedCornerShape(4.dp),
+                                    )
+                                    .padding(horizontal = 8.dp, vertical = 4.dp),
+                            ) {
+                                Text(
+                                    text = strengthLabel,
+                                    style = MaterialTheme.typography.labelSmall,
+                                    color = Color.White,
```
biplab1: Would it make sense to use `MaterialTheme.colorScheme` here instead of hardcoding the color? Might help with consistency and theming

core/ui/src/commonMain/kotlin/org/mifospay/core/ui/utils/PasswordChecker.kt
```diff
@@ -83,14 +74,11 @@ object PasswordChecker {
...
-        if (password.length < STRONG_PASSWORD_LENGTH) {
-            feedback.add("For a stronger password, use at least $STRONG_PASSWORD_LENGTH characters.")
-        }
         if (password.hasConsecutiveRepetitions()) {
-            feedback.add("Remove consecutive repeating characters.")
+            feedback.add("Avoid using consecutive repeated characters.")
         }
         if (password.hasSpaces()) {
-            feedback.add("Remove spaces.")
+            feedback.add("Do not include spaces in the password.")
```
biplab1, line 83: Since we’re targeting a global audience with our mobile wallet, wouldn’t it be better to use string resources instead of hardcoded strings? I noticed this usage of hardcoded strings in several places. Using `strings.xml` would make it much easier to manage translations later on.

core/ui/src/commonMain/kotlin/org/mifospay/core/ui/PasswordStrengthIndicator.kt
HekmatullahAmin: done

core/ui/src/commonMain/kotlin/org/mifospay/core/ui/utils/PasswordChecker.kt
HekmatullahAmin: @biplab1 for now, I'm not moving the password feedback messages to `string.xml`. The reason is that `getPasswordFeedback()` is called from the `SignUpViewModel`, and the result is stored in `SignUpState`. Since SignUpState is marked as `@Parcelize` (for SavedStateHandle support), it can't contain a List<StringResource>.

Also, I can't use `stringResource()` inside getPasswordFeedback() because it's not a composable function, and stringResource() is only accessible from composables.

We'll revisit this later once we refactor how localized strings are handled in view models. for now added a TODO

biplab1: Makes sense given the current limitations with `@Parcelize` and not being able to access string resources outside composables. 

Just sharing — I’ve used `org.jetbrains.compose.resources.getString` inside a coroutine in one of my modules to get localized strings, and it worked fine. Might be something we can explore during the refactor. 

Reference: [To get string resources as a String from non-composable code](https://www.jetbrains.com/help/kotlin-multiplatform-dev/compose-multiplatform-resources-usage.html#p91qgw_112)

revanthkumarJ (approved):

niyajali (approved):

PR #1873, opened 2025-06-26
refactor(core, feature): replaced hard-coded colors.

sam-arth07:
## Issue Fix
Fixes #MW-188
Jira Task: [MW-188](https://mifosforge.jira.com/browse/MW-188)

## Screenshots

Before:

https://github.com/user-attachments/assets/c7257520-74f5-438d-b654-eec17616eb99

---

After:

[After.webm](https://github.com/user-attachments/assets/497a9202-cc08-48b8-9297-c0a54c271a7b)

## Description
What was achieved in this PR:

- Centralized color usage through MaterialTheme references
- Fixed all core UI components & feature modules to use theme-based colors
- Replaced hardcoded hex values with semantic color tokens
- Eliminated Color.White, Color.Black, Color.Gray hardcoded usage

##

[MW-188]: https://mifosforge.jira.com/browse/MW-188?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/ui/src/commonMain/kotlin/org/mifospay/core/ui/ExpiryDateInput.kt
```diff
@@ -119,9 +122,9 @@ fun FormattedDateView(
         text = char,
         style = MaterialTheme.typography.headlineSmall,
         color = if (isFocused) {
-            Color.DarkGray
+            MaterialTheme.colorScheme.darkGray
         } else {
-            Color.LightGray
+            MaterialTheme.colorScheme.lightGray
```
biplab1: The colors `darkGray` and `lightGray` are still statically defined in the design system, so while they're accessed via `MaterialTheme.colorScheme`, this doesn't provide true Material theming or dynamic adaptation to theme changes.

core/designsystem/src/commonMain/kotlin/org/mifospay/core/designsystem/theme/Color.kt
```diff
@@ -165,6 +168,180 @@ val surfaceContainerDark = Color(0xFF1C2025)
...
+    get() = Color.DarkGray
+
+val ColorScheme.lightGray: Color
+    @Composable
+    get() = Color.LightGray
+
+val ColorScheme.gray: Color
+    @Composable
+    get() = Color.Gray
+
```
biplab1: Using a static color like `Color.Gray` as an extension on ColorScheme is technically valid, but it doesn't leverage theming—consider defining theme-aware colors in your color scheme for better adaptability.

core/ui/src/commonMain/kotlin/org/mifospay/core/ui/OtpTextField.kt
```diff
@@ -125,9 +126,9 @@ fun CharView(
         text = char,
         style = MaterialTheme.typography.headlineSmall,
         color = if (isFocused) {
-            Color.DarkGray
+            MaterialTheme.colorScheme.darkGray
         } else {
-            Color.LightGray
+            MaterialTheme.colorScheme.lightGray
```
biplab1: same here, please review

core/designsystem/src/commonMain/kotlin/org/mifospay/core/designsystem/component/BottomSheet.kt
```diff
@@ -54,7 +55,7 @@ fun MifosBottomSheet(
 
     AnimatedVisibility(visible = showBottomSheet) {
         ModalBottomSheet(
-            containerColor = Color.White,
+            containerColor = MaterialTheme.colorScheme.white,
```
biplab1: It doesn't actually serve the purpose of using MaterialTheme.colorscheme if you are still using hard coded colors in `common:designsystem`.

biplab1:
Here are some references you could follow: 
1. [Material Design 3](https://developer.android.com/develop/ui/compose/designsystems/material3)
2. [Now in Android app](https://github.com/android/nowinandroid/tree/main)
3. [KMP Project Template](https://github.com/openMF/kmp-project-template/tree/dev)

core/ui/src/commonMain/kotlin/org/mifospay/core/ui/ExpiryDateInput.kt
sam-arth07: I have removed the static usage and replaced it with Material Design system's theme adaptive colorscheme as you mentioned. Reused the already defined colors like `MaterialTheme.colorScheme.scrim` instead of `Color.Black` after carefully looking at the corresponding effect on the UI.

core/ui/src/commonMain/kotlin/org/mifospay/core/ui/OtpTextField.kt
sam-arth07: Replaced static colors lightGray and darkGray with MT's `outlineVariant` and `outline` resp.

core/designsystem/src/commonMain/kotlin/org/mifospay/core/designsystem/component/BottomSheet.kt
sam-arth07: Have removed all the previously defined hardcoded basic colors from `commom:designsystem`.  If there is scope for improvement do let me know.

biplab1:
@sam-arth07 If possible, could you please share screen recordings from before and after the refactor? It will help check that everything still looks good after the changes.

biplab1:
@sam-arth07 I’ve reviewed the changes — they look good to me for now. Some of the colors you’ve defined using extension functions in the `MaterialTheme` color scheme — please check with Niyaj once in case these need to be changed later. Also, if possible, please share screen recordings from before and after the refactor so we can make sure everything still looks fine.

core/designsystem/src/commonMain/kotlin/org/mifospay/core/designsystem/theme/Color.kt
```diff
@@ -165,6 +168,153 @@ val surfaceContainerDark = Color(0xFF1C2025)
...
+    @Composable
+    get() = MaterialTheme.colorScheme.surface.copy(
+        red = 0.94f,
+        green = 0.94f,
+        blue = 0.94f,
+    )
+
 // New UI Colors
 object NewUi {
     val walletColor1 = Color(0xFF1f7dd5)
```
niyajali: Also create extension variable of these colours

niyajali (changes requested):

core/designsystem/src/commonMain/kotlin/org/mifospay/core/designsystem/theme/Color.kt
```diff
@@ -165,6 +168,153 @@ val surfaceContainerDark = Color(0xFF1C2025)
 val surfaceContainerHighDark = Color(0xFF272A2F)
 val surfaceContainerHighestDark = Color(0xFF32353A)
 
+// Extension Colors for PasswordStrengthIndicator
+val ColorScheme.veryStrongPassword: Color
+    @Composable
```
niyajali: Can you use the similar M3 colors instead of these colours like primary, secondary, error, and tertiary based on the password strength also update below colours to m3 colors

core/designsystem/src/commonMain/kotlin/org/mifospay/core/designsystem/theme/Color.kt
```diff
@@ -165,6 +168,153 @@ val surfaceContainerDark = Color(0xFF1C2025)
...
+    get() = Color(0xFFFF8800)
+
+val ColorScheme.mediumPriority: Color
+    get() = Color(0xFFFFBB33)
+
+val ColorScheme.lowPriority: Color
+    get() = Color(0xFF99CC00)
+
+// Extension Colors for SavingAccountsStatusChips
+val ColorScheme.statusChipPendingApproval: Color get() = Color(0xFFFFF9C4)
```
niyajali: Here and below chips colour too update with similar M3 colors and make sure it reflects properly on both mode, basically I don't want to use any custom colors on these components the design might be little different but it will help future.

niyajali (changes requested):

biplab1 (approved):
The changes look good to me. I have reviewed the uploaded videos before and after fix. This can be merged once the remaining requested updates are implemented.

core/designsystem/src/commonMain/kotlin/org/mifospay/core/designsystem/theme/Color.kt
sam-arth07: Ok, however if we create extension variable using M3 color tokes like in place of `Color(0xFF1f7dd5)` we can use for eg, `MaterialTheme.colorScheme.primary `  then this will lead to change in the defined colors. Is this acceptable for the current implementation?

core/designsystem/src/commonMain/kotlin/org/mifospay/core/designsystem/theme/Color.kt
sam-arth07: Here too, like these colors are for the specific elements in the app, so can I change the previously defined colors to something else considering If there will be future UI mockups which suggests that the colorScheme might change so are these changes necessary? Also if we want to update with M3 color tokes like primary, secondary etc.. what colors do I decide for each element because the defined colors are quite different from ones used for M3. 
For eg, for strong password we are using : `Color(0xFF41B06D)` which is a greenish tone but there is no similar color in the M3 color tokens.

core/designsystem/src/commonMain/kotlin/org/mifospay/core/designsystem/theme/Color.kt
```diff
@@ -165,8 +168,199 @@ val surfaceContainerDark = Color(0xFF1C2025)
...
+    @Composable
+    get() = MaterialTheme.colorScheme.secondary
+
+val ColorScheme.lowPriority: Color
+    @Composable
+    get() = MaterialTheme.colorScheme.tertiary
+
+// Extension Colors for SavingAccountsStatusChips
+val ColorScheme.statusChipPendingApproval: Color
+    @Composable
```
niyajali: Just apply these colours directly to the specific components and no need to create extension function also remove all inline comments like `val pa= etc // inline comments remove this if needed move into top of the variable declaration`

niyajali (changes requested):

core/designsystem/src/commonMain/kotlin/org/mifospay/core/designsystem/theme/Color.kt
sam-arth07: @niyajali I have made the requested changes. Removed all the inline comments and applied the colors directly to the targeted components.

niyajali (approved):

PR #1876, opened 2025-06-29
Fix crash when app is opened without internet after passcode entry

HekmatullahAmin:
Jira Task: [MW-217](https://mifosforge.jira.com/browse/MW-217)

### Before

https://github.com/user-attachments/assets/a4897e83-4fe4-472a-b651-d367ac22071c

### After

https://github.com/user-attachments/assets/363f0e57-3abc-400f-9b3c-61819b3fe4e2

[MW-217]: https://mifosforge.jira.com/browse/MW-217?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/feature/home/src/commonMain/kotlin/org/mifospay/feature/home/HomeViewModel.kt
+++ b/feature/home/src/commonMain/kotlin/org/mifospay/feature/home/HomeViewModel.kt
@@ -11,7 +11,6 @@ package org.mifospay.feature.home
 
 import androidx.lifecycle.viewModelScope
 import kotlinx.coroutines.ExperimentalCoroutinesApi
-import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.SharingStarted
 import kotlinx.coroutines.flow.flatMapLatest
 import kotlinx.coroutines.flow.mapLatest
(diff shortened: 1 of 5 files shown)


HekmatullahAmin:
Previously, `getActiveAccountsWithTransactions()` was returning `DataState.Success` even when an exception occurred during API calls, because errors from the combined flow were not properly caught.
To address this, I replaced the manual `.map { DataState.Success(...) }` with a call to `asDataStateFlow()`, which ensures correct emission of Loading, Success, and Error states with proper exception handling.

I also left the hardcoded error string as-is for now, since we plan to migrate all strings to stringResource() via a separate localization ticket.

revanthkumarJ (approved):

feature/home/src/commonMain/kotlin/org/mifospay/feature/home/HomeViewModel.kt
```diff
@@ -44,10 +46,26 @@ class HomeViewModel(
...
+     *
+     * Changing the value causes the `flatMapLatest` block in `accountState` to re-emit,
+     * which triggers a fresh API call via `repository.getActiveAccountsWithTransactions`.
+     *
+     * This mechanism ensures re-fetching only occurs when explicitly triggered, and works
+     * well in combination with UI states like `ViewState.Error`, where the retry action is visible.
+     */
+    private val reloadTrigger = MutableStateFlow(false)
+
+    val accountState = reloadTrigger.flatMapLatest {
```
niyajali: This will reload on both cases(true|false)

niyajali: Use distinctUntilChanged or a map function to check only when reload is true. Alternatively, you can move the reloadTrigger parameter to HomeState and check it using stateFlow.map{ it.reloadTrigger }.flatMapLatest { } something like this, you may find similar implementation on other screen

niyajali (changes requested):

feature/home/src/commonMain/kotlin/org/mifospay/feature/home/HomeViewModel.kt
HekmatullahAmin: @niyajali ,

Thanks for your feedback! I understand your suggestion about limiting the trigger using `distinctUntilChanged()` or mapping based on a `true` flag, or even moving the trigger into the `HomeState`. That’s a valid pattern in many cases.

However, in this particular case, I intentionally implemented it as:
```
is HomeAction.OnRetryClicked -> {
    reloadTrigger.value = !reloadTrigger.value
}
```
…because I wanted the reload to fire every time the user clicks retry, regardless of the previous state — treating it like an event rather than a toggle with semantic meaning.

### Why This Approach
Imagine this sequence:

1. The user opens the app with no internet.

2. The HomeScreen triggers data fetch once (via `flatMapLatest`) and falls into `ViewState.Error`.

3. `ErrorScreenContent` becomes visible, showing a Retry button.

4. The user taps Retry → `reloadTrigger` toggles, fetch is retried.

- But if internet is still off, `Error` shows again.

5. Now the user turns on internet and taps Retry again.

- With my current toggle-based approach, the value changes again (e.g., `true → false`) → `flatMapLatest` is re-triggered, and data is fetched successfully.

If I hadn't done this , then:

- The first tap sets it to true.

- If the internet is still off → fetch fails.

- On the second tap, nothing happens, because the state (true) hasn’t changed, and StateFlow only emits on new values.

That would break the Retry experience unless we manually reset the trigger to false inside the Error state — which would again cause two state changes, potentially leading to an unnecessary double execution.
Yes, I could’ve added something like:
```
if (result is DataState.Error) {
    reloadTrigger.value = false
}
```
But that adds an implicit side effect inside mapLatest, which:

- Creates a second trigger (on error state)

- Can cause double emissions, especially during recompositions or lifecycle changes

### Consistency with Other Screens
I also checked the other usages of `ErrorScreenContent`:

- `SpecificTransactionScreen.kt` – line 88

- `InvoiceDetailsScreen.kt` – line 106

- `ProfileScreen.kt` – line 108

In all of these cases, the Retry button is displayed, but no ViewModel logic is wired up to handle it — so currently tapping Retry does nothing.

In contrast, in HomeViewModel I wanted to make the retry functional and user-driven, not state-driven.

niyajali: Okay, move that property(reloadTrigger) to HomeState param

niyajali: Also enable the PullRefresh Functionality

niyajali (changes requested):

HekmatullahAmin:
@niyajali the implementation of `pulltorefresh` is done.


https://github.com/user-attachments/assets/5795476e-53e2-4d6c-b6a3-ef408c84b789

feature/home/src/commonMain/kotlin/org/mifospay/feature/home/HomeViewModel.kt
```diff
@@ -147,6 +184,8 @@ class HomeViewModel(
 data class HomeState(
     val client: Client,
     val defaultAccountId: Long?,
+    val reloadTrigger: Boolean = false,
```
niyajali, line 187: I don't think we need this parameter we can use the isRefreshing parameter to handle the refresh/retry logic, and also we no need multiple retry/reload method we can use one .. basically it reload right on both cases.

niyajali (changes requested):

biplab1 (approved):
The changes look good to me and can be merged.

Minor suggestion:
Instead of toggling `reloadTrigger` as a Boolean, you could use an `Int` and increment it:

```kotlin
it.copy(reloadTrigger = it.reloadTrigger + 1)
```

This ensures each reload signal is unique, so rapid retries or duplicate updates aren’t ignored.

feature/home/src/commonMain/kotlin/org/mifospay/feature/home/HomeViewModel.kt
biplab1: Keeping `reloadTrigger` separate from `isRefreshing` can help clarify that `isRefreshing` is just for the UI spinner, while `reloadTrigger` is purely a reload signal. Combining them can work too, just mentioning it in case it helps with future clarity.

HekmatullahAmin: @niyajali,

Just to clarify  `isRefreshing` and `reloadTrigger` serve two separate purposes.

- isRefreshing is strictly for UI — it controls whether to show the refresh spinner (e.g., during pull-to-refresh).
- reloadTrigger is an event flag to actually trigger the reload logic.

Currently, I use two distinct actions:

- OnRetryClicked → reloads without showing the spinner
- OnPullToRefresh → reloads and shows the spinner

Even though both actions ultimately trigger the same data reload, the UX behavior differs — especially in whether isRefreshing should be true or not.

If we wanted to unify them into a `single action`, then we would need to pass down the `isRefreshing` flag from the UI layer:

- true for pull-to-refresh
- false when user taps Retry

But either way, we need both states to support the right user feedback.

Let me know if you'd prefer the unified-action approach — I can adapt, just wanted to make sure the distinction is clear.

niyajali:
@HekmatullahAmin it's actually reload right on both cases, and you can create multiple action on both cases but creating multiple parameters doesn't make sense to me as it does one job.

HekmatullahAmin:
@niyajali I think there was a bit of confusion let me clarify with example.

As I mentioned earlier, reloadTrigger is not a state, it's just an event flag that toggles between true and false each time the user performs an action (like Retry or Pull-to-Refresh). It’s used to trigger a reload via flatMapLatest, not to represent any UI state.
Below I  list two approaches

### Why we currently use two separate values (reloadTrigger and isRefreshing):
If we try to use only isRefreshing as both:

- a trigger for reloads,
- and a UI flag to show the refresh spinner,

…it causes unintended UX behavior.

Example:
- User taps the Retry button (not pull-to-refresh).
- We set isRefreshing = true to trigger the reload.
- Since the screen observes isRefreshing, it also shows the refresh spinner, which we don’t want for a Retry action by a Button.

So isRefreshing ends up doing two jobs (triggering logic and controlling UI).

That’s why I currently separate the concerns:

- reloadTrigger: triggers the data reload
- isRefreshing: strictly controls spinner visibility

### Alternative Approach (if we want to avoid using two values):
We can remove reloadTrigger altogether and instead:

- Introduce a separate function in the ViewModel to manually call:
```
repository.getActiveAccountsWithTransactions(...)
```
- Use that function in the Retry flow using the Button.

- For Pull-to-Refresh, we continue to use isRefreshing = true and trigger the existing accountState reload logic:
```
val accountState = stateFlow
    .mapLatest { it.reloadTrigger }
    .flatMapLatest { ... }
```
- Then reset isRefreshing to false on success or error as usual.

This would avoid the need for reloadTrigger, but introduce some duplication in how data is loaded.

Let me know which approach you’d prefer:

- Stick with current separation (reloadTrigger + isRefreshing) for clarity and simplicity
- Or refactor to the single-state + manual fetch method for retry cases

Happy to go either way — just wanted to clarify the reasoning.

niyajali:
Okay, got it, I've some different approach but no issues this will work so I'm merging this now.

niyajali (approved):

PR #1878, opened 2025-07-07
[MW-207] Fix iOS Navigation & Crash Issues, Add TransferDate Serializer, Improve Beneficiary & Home Screen Behavior

HekmatullahAmin:
Jira Task: [MW-207](https://mifosforge.jira.com/browse/MW-207)

This PR addresses several key issues related to app navigation and iOS-specific crashes, along with improvements in data handling and UI responsiveness.

## Fixes
- Resolved application crash on iOS 18 due to SavedStateHandle & Parcelable incompatibility.

- Fixed issue where deleted beneficiaries would not immediately disappear from the list. Now the UI updates correctly without requiring a screen reload or navigation.

- Implemented a custom serializer to handle inconsistent transferDate formats received from the server (String vs Array). This ensures proper parsing and avoids crashes.

Home screen cards now properly display details when tapped.

## Known Issue / Follow-up
The Share button functionality (which should trigger a bottom sheet dialog) is currently implemented for Android but not yet for iOS.

A separate ticket will cover the iOS implementation, as I'm currently facing issues with my macOS environment.

## Aftre video

https://github.com/user-attachments/assets/aedda9e2-a461-4b6b-8c3e-36671b9ac2e7

[MW-207]: https://mifosforge.jira.com/browse/MW-207?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/model/src/commonMain/kotlin/org/mifospay/core/model/savingsaccount/Currency.kt
```diff
@@ -23,4 +20,4 @@ data class Currency(
     val displaySymbol: String,
     val nameCode: String,
     val displayLabel: String,
-) : Parcelable
+)
```
biplab1, line 23: Adding default values to all optional `Currency` fields helps handle missing data safely. Additionally, providing defaults for `code` and `name` (which are required) ensures the app won’t crash if no currency is configured in the web app.

HekmatullahAmin: Hi @biplab1  ,
First off, thanks for pointing this out really appreciate it!
I actually had a similar thought initially about giving default values to some of the data classes like Currency, especially to avoid crashes if certain fields were missing from the API response.

But later I reconsidered — since our app is financial, having a missing or wrong currency could signal a serious bug. So I figured we should push the caller to explicitly provide that value, rather than fall back to a default and potentially mask the issue.

I was kind of thinking both ways, so for now I just kept it as it was before — strict with no defaults.
correct me if i am wrong.

core/model/src/commonMain/kotlin/org/mifospay/core/model/savingsaccount/SavingAccount.kt
```diff
@@ -9,10 +9,6 @@
  */
 package org.mifospay.core.model.savingsaccount
 
-import org.mifospay.core.common.Parcelable
-import org.mifospay.core.common.Parcelize
-
-@Parcelize
 data class SavingAccount(
```
Nagarjuna0033, line 12: @HekmatullahAmin don't we need serializable annotation for this data class, is it intentional or have you forgot?

HekmatullahAmin: @Nagarjuna0033 it was only used on `ClientAccount` which was not used anywhere

image

core/model/src/commonMain/kotlin/org/mifospay/core/model/savingsaccount/Currency.kt
biplab1: You’re right — since this is a financial app, all financial products depend on having at least one valid currency configured. So it makes sense to keep `code` and `name` without defaults, because the backend can’t really function without at least one currency defined.

For the other fields, though, these are optional — as you can see in the [web-app implementation](https://github.com/openMF/web-app/blob/dev/src/app/shared/models/general.model.ts). Adding sensible defaults for these helps prevent crashes if any values are missing in the response.

revanthkumarJ (approved):

Nagarjuna0033 (approved):

biplab1 (approved):
The changes are fine from my side, good to go.

PR #1883, opened 2025-07-16
refactor(feature:send-money): replace hardcoded strings with string resource

biplab1:
## Issue Fix
Jira Task: [MW-231](https://mifosforge.jira.com/browse/MW-231)

## Description
   - Replaced hardcoded strings with string resources
   - Added dialog state to handle generic errors using string resources
   
These changes improve support for language localization.
 
<p float="left">
  
  
</p>

##

[MW-231]: https://mifosforge.jira.com/browse/MW-231?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/feature/send-money/src/commonMain/kotlin/org/mifospay/feature/send/money/SendMoneyScreen.kt
+++ b/feature/send-money/src/commonMain/kotlin/org/mifospay/feature/send/money/SendMoneyScreen.kt
@@ -56,7 +56,18 @@ import androidx.compose.ui.unit.dp
 import androidx.lifecycle.compose.collectAsStateWithLifecycle
 import mobile_wallet.feature.send_money.generated.resources.Res
 import mobile_wallet.feature.send_money.generated.resources.feature_send_money_amount
+import mobile_wallet.feature.send_money.generated.resources.feature_send_money_bottom_bar
+import mobile_wallet.feature.send_money.generated.resources.feature_send_money_close
+import mobile_wallet.feature.send_money.generated.resources.feature_send_money_loading
+import mobile_wallet.feature.send_money.generated.resources.feature_send_money_no_accounts_found
+import mobile_wallet.feature.send_money.generated.resources.feature_send_money_oops
+import mobile_wallet.feature.send_money.generated.resources.feature_send_money_proceed
+import mobile_wallet.feature.send_money.generated.resources.feature_send_money_scan_qr
+import mobile_wallet.feature.send_money.generated.resources.feature_send_money_selected
 import mobile_wallet.feature.send_money.generated.resources.feature_send_money_send
+import mobile_wallet.feature.send_money.generated.resources.feature_send_money_something_went_wrong
+import mobile_wallet.feature.send_money.generated.resources.feature_send_money_to_account
+import mobile_wallet.feature.send_money.generated.resources.feature_send_money_vpa_mobile_account_number
 import org.jetbrains.compose.resources.stringResource
 import org.koin.compose.viewmodel.koinViewModel
 import org.mifospay.core.common.utils.maskString
@@ -149,7 +160,7 @@ private fun SendMoneyScreen(
                             ) {
                                 Icon(
                                     imageVector = MifosIcons.Scan,
-                                    contentDescription = "Scan QR",
+                                    contentDescription = stringResource(Res.string.feature_send_money_scan_qr),
                                 )
                             }
                         },
@@ -216,7 +227,7 @@ private fun SendMoneyBottomBar(
         ) {
             AnimatedVisibility(
                 visible = showDetails && selectedAccount != null,
-                label = "BottomBar",
+                label = stringResource(Res.string.feature_send_money_bottom_bar),
                 enter = fadeIn() + slideInVertically(
                     initialOffsetY = { fullHeight ->
                         fullHeight / 4
@@ -242,7 +253,7 @@ private fun SendMoneyBottomBar(
                 enabled = showDetails,
                 modifier = Modifier.fillMaxWidth(),
             ) {
-                Text(text = "Proceed")
+                Text(text = stringResource(Res.string.feature_send_money_proceed))
             }
         }
     }
@@ -260,7 +271,7 @@ private fun SelectedAccountCard(
         verticalArrangement = Arrangement.spacedBy(8.dp),
     ) {
         Text(
-            text = "To Account",
+            text = stringResource(Res.string.feature_send_money_to_account),
             style = MaterialTheme.typography.labelLarge,
         )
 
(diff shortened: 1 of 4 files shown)


feature/send-money/src/commonMain/kotlin/org/mifospay/feature/send/money/SendMoneyViewModel.kt
```diff
@@ -253,3 +256,10 @@ sealed interface SendMoneyAction {
 
     data class HandleRequestData(val requestData: String) : SendMoneyAction
 }
+
+object Strings {
+    const val ERROR_AMOUNT_EMPTY = "error_amount_empty"
```
revanthkumarJ: i think we can use enum here instead of object.

I assume the values like "error_amount_empty" not used anywhere

enum class Strings {
    ERROR_AMOUNT_EMPTY,
    ERROR_INVALID_AMOUNT
}

feature/send-money/src/commonMain/kotlin/org/mifospay/feature/send/money/SendMoneyViewModel.kt
```diff
@@ -216,7 +216,10 @@ data class SendMoneyState(
         data object Loading : DialogState
 
         @Serializable
-        data class Error(val message: String) : DialogState
+        data class Error(
+            val key: String,
+            val args: List<String> = emptyList(),
```
revanthkumarJ: why args any reason?

I think we pass StringResource and use it in ui

revanthkumarJ: I saw in ui you only used key

revanthkumarJ (changes requested):

feature/send-money/src/commonMain/kotlin/org/mifospay/feature/send/money/SendMoneyViewModel.kt
```diff
@@ -138,11 +138,11 @@ class SendMoneyViewModel(
     }
 
     private fun validateTransferFlow() = when {
-        state.amount.isBlank() -> updateErrorState("Amount cannot be empty")
+        state.amount.isBlank() -> updateErrorState(Strings.ERROR_AMOUNT_EMPTY)
 
-        state.amount.toDoubleOrNull() == null -> updateErrorState("Invalid amount")
+        state.amount.toDoubleOrNull() == null -> updateErrorState(Strings.ERROR_INVALID_AMOUNT)
```
revanthkumarJ, line 153: I don't know why you userd String and passing keys and mapping StringResources in ui

Based on previously implementation they passed strings, now we can replace them with StringResources and show them on ui.

If any specific reason for doing so can you mention it

revanthkumarJ (changes requested):

feature/send-money/src/commonMain/kotlin/org/mifospay/feature/send/money/SendMoneyViewModel.kt
biplab1: I have reverted it back to its original state. Thanks!

feature/send-money/src/commonMain/kotlin/org/mifospay/feature/send/money/SendMoneyViewModel.kt
biplab1: We can’t use `stringResource` here since this function is not a `@Composable`, and we can’t use `getString` either because it must be called inside a `suspend` function.

feature/send-money/src/commonMain/kotlin/org/mifospay/feature/send/money/SendMoneyViewModel.kt
biplab1: `updateErrorState()` requires a `String` to be passed, which is then used in `DialogState.Error`, which also accepts a `String` as its argument. So I used an `object` with `const val` strings to keep it simple and consistent.

If I switch to using an enum instead, I’d also need to update `Error` to accept the enum type and handle mapping the QR error (and any other errors with arguments, which I had done earlier but reverted based on your other comment) separately to allow resolving the arguments in the UI layer.

biplab1: Currently, I have used `getString` for QR error inside `Error` which works because it is called inside `viewModelScope.launch`.

feature/send-money/src/commonMain/kotlin/org/mifospay/feature/send/money/SendMoneyViewModel.kt
biplab1: If we want `updateErrorState` to accept `StringResource`, then we’d also need to update `Error`, but `StringResource` is not serializable.

feature/send-money/src/commonMain/kotlin/org/mifospay/feature/send/money/SendMoneyViewModel.kt
biplab1: The initial idea was to use an `object` for `Strings` and map the keys to `stringResource` in the UI. `Error` was modified to use arguments since the invalid QR message required them, but since we can use `getString` inside `Error` for the invalid QR (called within `viewModelScope.launch`), I reverted that change.

niyajali (changes requested):
@biplab1 take a look at the change password PR(#1885), add one more dialog state to show the generic messages using stringResource and keep the existing error state to show the server side errors.

biplab1:
@niyajali Thanks for the suggestion! I have added a dialog state to handle generic messages using string resources and kept the dialog error state as it is for server side errors. Let me know if anything else is required.

revanthkumarJ (approved):

feature/send-money/src/commonMain/kotlin/org/mifospay/feature/send/money/SendMoneyViewModel.kt
```diff
@@ -217,6 +235,8 @@ data class SendMoneyState(
 
         @Serializable
         data class Error(val message: String) : DialogState
+
+        data class ValidationError(val res: StringResource) : DialogState
```
niyajali, line 239: Annotate with `@Contextual`

feature/send-money/src/commonMain/kotlin/org/mifospay/feature/send/money/SendMoneyScreen.kt
```diff
@@ -506,6 +517,13 @@ private fun SendMoneyDialogs(
             visibilityState = LoadingDialogState.Shown,
         )
 
+        is SendMoneyState.DialogState.ValidationError -> MifosBasicDialog(
```
niyajali, line 520: Okay, I'll approve this, but try to use the same naming convention as in others for better consistency. We need to find a way to display the Server error in a human-readable format, which would make refactoring easier.

niyajali (approved):

PR #1884, opened 2025-07-16
feat(feature:ui): implement iOS file sharing and improve Android logic

HekmatullahAmin:
## Issue Fix
Jira Task: [MW-235](https://mifosforge.jira.com/browse/MW-235)

## What’s in this PR
This PR introduces a complete platform implementation for file sharing on both iOS and Android, as well as some related cleanup and improvements across the codebase.

### iOS Implementation
- Added actual implementation of ShareUtils.shareFile() for iOS using `FileKit.shareFile`.
- Supports sharing files (PDF, IMAGE) , TEXT file types.
- Compression support for images using the new `FileKit.compressImage` API.

### Android Enhancements
- Previous Android logic only supported image sharing.
- Now supports sharing other file types like PDFs, which is particularly useful in screens like `Transaction Details`, where the user can now share receipts or summaries in PDF format.
- Updated sharing logic to write any file to the cache directory.

### Library Update
- Bumped `FileKit` from `0.8.7` to `0.10.0-beta04`.
- Removed usages of `file.supportsStreams()` which no longer exists in the new version.
- Used `FileKit.compressImage()` for both Android and iOS — supported platforms only.

### Cross-Platform Support
Since the share sheet dialog is only applicable to iOS and Android, alternative strategies were used for other platforms:

- On wasmJs and js, implemented file downloading using `FileKit.download`.
- On desktop, implemented file saving using `FileKit.openFileSaver`.

These additions ensure consistent file sharing/downloading functionality across supported targets.

## Android & IOS
https://github.com/user-attachments/assets/deb7113e-c360-4913-a739-d2cfc5bcc9b1

## Desktop

https://github.com/user-attachments/assets/bbb930b0-609c-4276-8d7c-72317fcabb17

## Js

https://github.com/user-attachments/assets/7b08f160-8147-46c2-a327-a5e065e1011b

[MW-235]: https://mifosforge.jira.com/browse/MW-235?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


HekmatullahAmin:
wanted to use `get<CoroutineDispatcher>(named(MifosDispatchers.IO.name))` but in `native` and `jvm` the `Default dispatcher` is used which i believe it should be `IO` the same as it is in android. so i used directly the IO dispatcher here

niyajali:
Use the FileKit library to share on the supported platforms and on remaining platforms download those shareable file, don't make it complicated

HekmatullahAmin:
> Use the FileKit library to share on the supported platforms and on remaining platforms download those shareable file, don't make it complicated

@niyajali ,

I first tried using `FileKit.shareFile` for the iOS implementation before your suggestion, but I ran into this error during testing: you can have a look at the image which didn't allow me to share.
```
Cannot issue sandbox extension for URL:/Users/hekmatullahamin/Library/Developer/CoreSimulator/Devices/858235FC-8B91-44A2-BF75-4F1A262F7529/data/Containers/Data/.../code.png
```
This seems to be a sandbox-related issue in the iOS simulator — the system was unable to grant the necessary file access for sharing, likely because the file was outside of an allowed location or wasn’t handled in a way UIActivityViewController expects.

Then cause of that i used the native implementation. But after your suggestion i went and tried again.
To fix this, I created the NSURL manually using:
```
val nsUrl = NSURL.fileURLWithPath(tempFile.absolutePath())
```
That resolved the issue, and sharing works properly now using the native controller.

For the other platforms:

- I used `FileKit.download` for `wasmJs` and `js`, since that's the only supported option there.
- On desktop, I used `FileKit.openFileSaver`, because neither `FileKit.shareFile` nor `FileKit.download` is available for the JVM target.

Screenshot 2025-07-17 at 09 24 14

feature/kyc/src/commonMain/kotlin/org/mifospay/feature/kyc/KYCLevel2Screen.kt
```diff
@@ -203,36 +204,12 @@ private fun DocumentPicker(
     )
 
     val filePicker = rememberFilePickerLauncher(
-        mode = PickerMode.Single,
+        mode = FileKitMode.Single,
```
niyajali: Can't we do something like this

```
            is RegistrationUiAction.UpdateLogo -> {
                viewModelScope.launch {
                    val data = FileKit.openFilePicker(FileKitType.Image)

                    data?.let { file ->
                        mutableStateFlow.update {
                            it.copy(
                                logo = file.toImageBitmap().encodeToByteArray().toBase64DataUri(),
                            )
                        }
                    }
                }
            }
```

niyajali (changes requested):
Take a look into this for reference - https://github.com/openMF/kmp-project-template/blob/dev/core-base/ui/src/jsCommonMain/kotlin/template/core/base/ui/ShareUtils.kt

feature/kyc/src/commonMain/kotlin/org/mifospay/feature/kyc/KYCLevel2Screen.kt
HekmatullahAmin: @niyajali  both way is possible since we are in composable thin i believe using the composable is better here.
image
image

niyajali: Yes, I understand we can handle it in the composable function, but since we're managing everything in the ViewModel and I prefer not to handle it here, let's stay consistent. When selecting the file, you're already passing it to the ViewModel for processing, so we should handle everything within the ViewModel to maintain our current architecture.

HekmatullahAmin: @niyajali I moved the logic into viewmodel. And for `js` and `wasmJs` i used the download file and for `desktop` i used `openFileSaver` and i didn't use the `saveImageToGallery` cause we want to save file not just image.

HekmatullahAmin: So right now the `ShareUtils` shareFile method cover sharing any type of file whether image or pdf cause we have other file to share in our projects not just iamge to share

niyajali (approved):

PR #1896, opened 2025-07-30
refactor(core): replace MaterialTheme with KptMaterialTheme

HekmatullahAmin:
## Issue Fix
Fixes #{Issue Number}
Jira Task: [MW-245](https://mifosforge.jira.com/browse/MW-245)

[MW-245]: https://mifosforge.jira.com/browse/MW-245?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


HekmatullahAmin:
@niyajali ,
Just wanted to give you a quick heads-up on how I’ve used `KptTheme.spacing` and `shapes`:

I’ve applied KptTheme.spacing mainly for paddings and Spacer components.

For example, if a value was around 2.dp, I used xs; for something like 20.dp, I used lg — basically choosing the closest token within that range.

Spacing Token Mapping by Value Range
- Values > 0 and ≤ 4.dp → used xs (4.dp)
- Values > 4.dp and ≤ 8.dp → used sm (8.dp)
- Values > 8.dp and ≤ 16.dp → used md (16.dp)
- Values > 16.dp and ≤ 24.dp → used lg (24.dp)
- Values > 24.dp and ≤ 32.dp → used xl (32.dp)
- Values > 32.dp → used xxl (64.dp)

Same logic goes for shapes — I used `KptTheme.shapes.small`, medium, etc., depending on the closest matching corner radius.

However, I didn’t use spacing tokens for things like `.size()`, `.width()`, or `.height()`, since those sometimes involved fixed dimensions like 100.dp or 200.dp — which the design tokens don’t currently cover.

Also, regarding extension functions like:
```
KptTheme.spacing.paddingValues(start = ..., top = ..., ...)
```
I chose not to use those, mainly because I wanted precise control and to stick with the dimensions we’ve been using earlier in the project. For example, instead of using the full-edge paddingValues, I did:
```
.padding(start = KptTheme.spacing.lg, end = KptTheme.spacing.lg)
```
since the extension would apply padding to all sides, which wasn't always needed.

core/designsystem/src/commonMain/kotlin/org/mifospay/core/designsystem/theme/Type.kt
```diff
@@ -43,46 +43,38 @@ private fun fontFamily(): FontFamily {
     )
 }
 
-// Set of Material typography styles to start with
-@Composable
-internal fun mifosTypography() = Typography(
+internal val mifosTypography = KptTypographyImpl(
     displayLarge = TextStyle(
-        fontFamily = fontFamily(),
```
niyajali, line 50: Don't remove this parameter from here..so you no need overwrite that value in theme.kt file

core/designsystem/src/commonMain/kotlin/org/mifospay/core/designsystem/theme/Theme.kt
```diff
@@ -193,15 +194,67 @@ fun MifosTheme(
...
         LocalTintTheme provides tintTheme,
     ) {
-        MaterialTheme(
-            colorScheme = colorScheme,
-            typography = mifosTypography(),
+        KptMaterialTheme(
+            darkTheme = darkTheme,
+            themeBuilder = { isDark ->
+                kptTheme {
+                    colors {
```
niyajali: Create a separate file for the color configuration and pass theme provider implementation using KptColorSchemeImpl class and use it directly on KptTheme class instead of dsl builder function which make it more readable and cleaner

core/designsystem/src/commonMain/kotlin/org/mifospay/core/designsystem/theme/Theme.kt
```diff
@@ -193,15 +194,67 @@ fun MifosTheme(
...
+                        background = selectedColorScheme.background
+                        onBackground = selectedColorScheme.onBackground
+                        surface = selectedColorScheme.surface
+                        onSurface = selectedColorScheme.onSurface
+                        surfaceVariant = selectedColorScheme.surfaceVariant
+                        onSurfaceVariant = selectedColorScheme.onSurfaceVariant
+                        outline = selectedColorScheme.outline
+                        outlineVariant = selectedColorScheme.outlineVariant
+                    }
+                    typography {
```
niyajali: Similarly here

niyajali (changes requested):

niyajali:
> @niyajali , Just wanted to give you a quick heads-up on how I’ve used `KptTheme.spacing` and `shapes`:
> 
> I’ve applied KptTheme.spacing mainly for paddings and Spacer components.
> 
> For example, if a value was around 2.dp, I used xs; for something like 20.dp, I used lg — basically choosing the closest token within that range.
> 
> Spacing Token Mapping by Value Range
> 
> * Values > 0 and ≤ 4.dp → used xs (4.dp)
> * Values > 4.dp and ≤ 8.dp → used sm (8.dp)
> * Values > 8.dp and ≤ 16.dp → used md (16.dp)
> * Values > 16.dp and ≤ 24.dp → used lg (24.dp)
> * Values > 24.dp and ≤ 32.dp → used xl (32.dp)
> * Values > 32.dp → used xxl (64.dp)
> 
> Same logic goes for shapes — I used `KptTheme.shapes.small`, medium, etc., depending on the closest matching corner radius.
> 
> However, I didn’t use spacing tokens for things like `.size()`, `.width()`, or `.height()`, since those sometimes involved fixed dimensions like 100.dp or 200.dp — which the design tokens don’t currently cover.
> 
> Also, regarding extension functions like:
> 
> ```
> KptTheme.spacing.paddingValues(start = ..., top = ..., ...)
> ```
> 
> I chose not to use those, mainly because I wanted precise control and to stick with the dimensions we’ve been using earlier in the project. For example, instead of using the full-edge paddingValues, I did:
> 
> ```
> .padding(start = KptTheme.spacing.lg, end = KptTheme.spacing.lg)
> ```
> 
> since the extension would apply padding to all sides, which wasn't always needed.

Those are only for example usages you create your own extension function based on your use case or needs

And if the default values aren't suitable then you can customise it and pass your own spacing, padding, elevation values like the theme and typography.

HekmatullahAmin:
@niyajali,
The thing is, our project uses a wider range of dp sizes, and if I overwrite the existing spacing values directly, I risk unintentionally affecting parts of the app that already rely on the current values.

What I initially wanted to do was something like this:
```
interface ExtendedKptSpacing : KptSpacing {
    val xsLow: Dp
    val xsHigh: Dp
    ...
    val xxxl: Dp
}
```
With an ExtendedKptSpacingImpl that adds low/high variants between tokens like xs, sm, md, etc. So for example:

xs = 4.dp, but xsLow = 2.dp and xsHigh = 6.dp

sm = 10.dp, smLow = 8.dp, smHigh = 12.dp

And so on...

Then I could do:
```
(KptTheme.spacing as ExtendedKptSpacing).xsHigh
```
However, since `core-base` is synced from our kmp-project-template, and it shouldn't be modified directly, I avoided extending it with these changes.

I also considered creating a separate singleton like for ex: `CustomTheme` the same as `KmptTheme` that exposes ExtendedSpacing, but that would mean updating all usages to use CustomTheme.spacing instead of KptTheme.spacing.

So instead, I stuck with using the closest existing token (xs, sm, md, etc.) based on the range, rather than forcing precise spacing.

Let me know if you think there’s a better way we could integrate extended tokens without diverging from the base. or this is fine

niyajali (approved):

PR #1898, opened 2025-08-02
refactor(core-network): integrate core-base/network to core/network

HekmatullahAmin:
## Issue Fix
Fixes #{Issue Number}
Jira Task: [MW-247](https://mifosforge.jira.com/browse/MW-247)

[MW-247]: https://mifosforge.jira.com/browse/MW-247?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


HekmatullahAmin:
@niyajali,

We should enhance the network layer to support proper response validation. Right now, we're missing a unified way to handle HTTP status codes and extract meaningful error messages from server responses to show to the user.

Ktor provides built-in [client response validation](https://ktor.io/docs/client-response-validation.html) that we can leverage. We can plug into HttpResponseValidator to:

- Handle both success (2xx) and error (non-2xx) responses uniformly

- Parse and expose server error messages in a consistent format

- Optionally map status codes to user-friendly error types or UI messages

PR #1899, opened 2025-08-04
fix(feature): fix DialogState serialization error

biplab1:
## Issue Fix
Jira Task: [270](https://mifosforge.jira.com/browse/MW-270)

## Screenshots

| Description Empty | Amount Empty |
|-------------------|--------------|
|  |  |

Screen recording hows the screens related to the view models affected by the change.

https://github.com/user-attachments/assets/5364bf88-ddbd-4ef8-885a-76eb508021ae

## Description
Summary of changes:
 - Fixed DialogState serialization by adding a custom string resource serializer

##


biplab1:
@niyajali Could you please have a look when you get a chance? If everything looks good, kindly go ahead and merge it.

cmp-shared/cmp_shared.podspec
```diff
@@ -50,5 +50,5 @@ Pod::Spec.new do |spec|
             SCRIPT
         }
     ]
-    spec.resources = ['build/compose/cocoapods/compose-resources']
+    spec.resources = ['build\compose\cocoapods\compose-resources']
```
HekmatullahAmin: @biplab1 this should be forwad slash. if it is generated by itself no worries you can rebase your branch after my PR is merged it will get fixed.

biplab1: Got it. That was auto-generated—I’ve reverted the change and pushed the update.

HekmatullahAmin (approved):

HekmatullahAmin:
@biplab1 update this so it can be merged

feature/home/src/commonMain/kotlin/org/mifospay/feature/home/HomeViewModel.kt
```diff
@@ -190,16 +191,13 @@ data class HomeState(
     val defaultAccountId: Long?,
     val reloadTrigger: Boolean = false,
     val isRefreshing: Boolean = false,
-    val dialogState: DialogState? = null,
+    @Transient val dialogState: DialogState? = null,
```
niyajali: No, need to apply this annotation and it could be serializable, as this only contains basic data types

feature/home/src/commonMain/kotlin/org/mifospay/feature/home/HomeViewModel.kt
```diff
@@ -190,16 +191,13 @@ data class HomeState(
...
 ) {
 
-    @Serializable
     sealed class DialogState {
 
-        @Serializable
         data object Loading : DialogState()
 
-        @Serializable
```
niyajali: Mark this as serializable

niyajali (changes requested):

niyajali:
We need to a find a solution or need to create our own serializable type for StringResource type and we can't mark all dialog state as transient in all screens in some cases we can make exceptions but not in every place

biplab1:
> We need to a find a solution or need to create our own serializable type for StringResource type and we can't mark all dialog state as transient in all screens in some cases we can make exceptions but not in every place

@niyajali Thanks for the review. I wanted to share a quick recap of the issue and a clarification ([Slack thread](https://mifos.slack.com/archives/C5TTJS4BX/p1754317045933889?thread_ts=1754306869.916079&cid=C5TTJS4BX)):

#### Context:

In `SendMoneyState`, we use `dialogState: DialogState?`, and the entire state is persisted via:

```kotlin
savedStateHandle.setSerialized(key = KEY_STATE, value = it)
```

Initially, `DialogState.Error.ResourceMessage` used `StringResource`, which caused a serialization error:

```
StringResource is not serializable
```

#### What I tried:

* Replaced `StringResource` with `SendMoneyErrorMessageKey` enum.
* Resolved the actual string in the UI using enum keys.
* Even after this, the following error persisted:

```
Serializer for subclass 'ResourceMessage' is not found in the polymorphic scope of 'DialogState'.
```

Even though:

* `DialogState` is a sealed interface
* Subclasses like `ResourceMessage` are annotated with `@Serializable`
* Enum is used instead of `StringResource`

#### What works:

Marking `dialogState` as `@Transient` in `SendMoneyState` prevents serialization errors and the error message still displays correctly in the UI.

```kotlin
@Transient val dialogState: DialogState? = null
```

* Also, tried just making the dialogState `@Transient` only in `SendMoneyViewModel.kt`, still the above error persisted. So, this approach had to be applied in other places using `DialogState` as well, such as `EditProfileViewModel.kt`, `HomeViewModel.kt`, and `MakeTransferViewModel.kt`.

---

### Question:

Why do we need `DialogState` to be serializable?

Since the UI shows the appropriate error messages even without serialization, is there a specific reason (e.g. state restoration after process death) to persist `dialogState`? If not, does it make sense to treat it as non-persistable UI state?

If needed, I can explore writing a custom serializable wrapper, but wanted to confirm the necessity first.

---

Let me know your thoughts.

feature/home/src/commonMain/kotlin/org/mifospay/feature/home/HomeViewModel.kt
biplab1: I have reverted the changes. Thanks!

feature/home/src/commonMain/kotlin/org/mifospay/feature/home/HomeViewModel.kt
biplab1: Marked this as `@Serializable`. Thanks!

biplab1:
@niyajali I have implemented the requested changes. Please have a look and if everything looks good, kindly merge it.

biplab1:
@niyajali Please have a look, I have used a custom serializer for string resource. It seems to work. 

https://github.com/user-attachments/assets/92f9c635-8478-4028-b558-8de1f4d42020

biplab1:
@niyajali A custom serializer has been added for handling string resources. Kindly review it.

https://github.com/user-attachments/assets/83e580e5-2d10-4139-888b-8f0673e9a1a3

biplab1:
@niyajali Other `@Transient` usages for `DialogStates` with string resources weren’t refactored, as I’d need to test all screens and don’t fully know their use cases. I focused on modules relevant to my C4GT work, but can handle the rest later if needed.

niyajali:
@biplab1 Thanks for the info. I was unwell and couldn't reply sooner, but I'll make sure to look into this soon.

niyajali (approved):

PR #1912, opened 2025-08-28
fix(home, history): see all; sorting; empty state; per-tab scroll; top bar

biplab1:
Jira Task: [298](https://mifosforge.jira.com/browse/MW-298)

## Screenshots

Before fix (state issue):

https://github.com/user-attachments/assets/1f483d6d-65ab-49a9-8a3e-fee093c1b67c

After fix (state and see all on click navigation):

https://github.com/user-attachments/assets/ec6e5955-f80d-49d4-a430-b927bd92e755

After fix (added history top bar):

https://github.com/user-attachments/assets/80cee731-a3f5-4422-b2bf-5becd3ab5958

After fix (transaction history sorted in descending order and tabs have independent scrolling)

https://github.com/user-attachments/assets/1dcfba88-33bb-457a-a4fa-b3d2a51628a9

## Description
  - Fixed the `see all` onclick navigation
  - Fixed empty state issue before loading history
  - Added history top bar
  - Added independent scrolling support for each tab
  - Sorted transactions by date in descending order

##


core/common/src/commonMain/kotlin/org/mifospay/core/common/DateHelper.kt
```diff
@@ -370,4 +369,44 @@ object DateHelper {
...
+            val monthName = parts[1]
+            val year = parts[2].toInt()
+
+            val month = when (monthName) {
+                "Jan" -> 1
+                "Feb" -> 2
+                "Mar" -> 3
+                "Apr" -> 4
+                "May" -> 5
+                "Jun" -> 6
```
HekmatullahAmin: we have pre defined public and private functions that either return the month numbers or data as `dd MMM yyyy`. so use them.
and since you are returning null in this case use runCatching and chain it with `getOrNull` it will returns the value if success, or null if it failed. :
```
return candidates.firstNotNullOfOrNull { fmt ->
        runCatching {
            val localDate = fmt.parse(cleaned)
            localDate.atStartOfDayIn(timeZone).toEpochMilliseconds()
        }.getOrNull()
    }
```

core/data/src/commonMain/kotlin/org/mifospay/core/data/repositoryImpl/SelfServiceRepositoryImpl.kt
```diff
@@ -156,7 +159,9 @@ class SelfServiceRepositoryImpl(
         return accountId.asFlow().flatMapMerge { clientId ->
             getSelfAccountTransactions(clientId)
         }.scan(emptyList()) { acc, transactions ->
-            acc + transactions.sortedByDescending { it.date }.let { sortedList ->
+            (acc + transactions).sortedByDescending { transaction ->
+                DateHelper.parseDateToMillis(transaction.date) ?: 0L
```
HekmatullahAmin: Using `0L` as a fallback isn’t ideal, because `0L` corresponds to a real date (1970-01-01 00:00:00 UTC). That means invalid or unparsable dates get treated the same as legitimate epoch transactions, which makes sorting ambiguous.

Instead, prefer:
```
DateHelper.parseDateToMillis(transaction.date) ?: Long.MIN_VALUE
```

Long.MIN_VALUE is a safe and is far, far earlier than any real date you’d ever encounter, so any invalid dates will always sort to the end of the list (when sorting descending).

HekmatullahAmin: - Example 1 — real epoch transaction

Suppose you actually have a transaction from 1 Jan 1970 (some systems do have old or test data like this).

Its timestamp = 0L.

Now suppose another transaction had a broken date string → parser fails → also gets 0L.

Both are indistinguishable. You cannot tell which one was real vs invalid.

- Example 2 — sorting confusion
Imagine you have three transactions:


Transaction1  - `2024-09-01` - 1725148800000
T2 - `bad-date` - null → fallback 0L
T3 - 1970-01-01` - 0L (real epoch)

Now you sort descending:

T1 (2024-09-01 → big millis value)

T2 ("bad-date" → 0L)

T3 (real 1970 → 0L)

See the problem?

T2 (invalid) and T3 (valid) are treated the same.

The “bad” one doesn’t sink to the bottom. Instead, it sits at the same level as real epoch data.

core/data/src/commonMain/kotlin/org/mifospay/core/data/repositoryImpl/SelfServiceRepositoryImpl.kt
```diff
@@ -167,16 +172,18 @@ class SelfServiceRepositoryImpl(
...
+            .flatMapLatest { accountIds ->
+                if (accountIds.isEmpty()) {
+                    flowOf(emptyList())
+                } else {
+                    getTransactions(accountId = accountIds, null)
+                        .filter { transactions -> transactions.isNotEmpty() }
+                }
             }
+            .asDataStateFlow()
             .flowOn(dispatcher)
```
HekmatullahAmin: - i think we already discussed for flow there is no need to again have a `flowOn`. 
- and also i think just adding the `asDataStateFlow` will take care of everything there wouldn't be a need for if/else block cause `getTransactions` will return empty list in case there is nothing just check if it works the remove the if/else and just use `getTransactions(accountId = accountIds, null)` and chain it with `asDataStateFlow()`

feature/history/src/commonMain/kotlin/org/mifospay/feature/history/HistoryScreen.kt
```diff
@@ -65,42 +72,57 @@ internal fun HistoryScreenContent(
     state: HistoryState,
     modifier: Modifier = Modifier,
     onAction: (HistoryAction) -> Unit,
+    showTopBar: Boolean = true,
+    onBackClick: (() -> Unit)? = null,
 ) {
-    Box(
+    Column(
```
HekmatullahAmin: use scaffold instead of column. it gives u a more consistent layout structure:

- Places the top bar in the correct position.
- Manages insets (system bars, status bar).
- Lets you easily add snackbars, floating action buttons, or bottom bars in the future.

HekmatullahAmin: and use the already created scaffolds

feature/home/src/commonMain/kotlin/org/mifospay/feature/home/HomeScreen.kt
```diff
@@ -146,7 +147,7 @@ internal fun HomeScreen(
                 navigateToTransactionDetail(event.accountId, event.transactionId)
             }
 
-            is HomeEvent.NavigateToTransactionScreen -> {}
+            is HomeEvent.NavigateToTransactionScreen -> navigateToHistory.invoke()
```
HekmatullahAmin: why u calling invoke on it just directly use navigateToHistory()

biplab1: In this file, the other functions are also invoked in the same way, so I kept it consistent.

feature/history/src/commonMain/kotlin/org/mifospay/feature/history/HistoryScreen.kt
biplab1: Refactored to use `MifosScaffold` and simplified padding in each view state; the Box was only used in the loading state to avoid covering the whole screen.

core/data/src/commonMain/kotlin/org/mifospay/core/data/repositoryImpl/SelfServiceRepositoryImpl.kt
biplab1: Updated to use `Long.MIN_VALUE`. 1970 transactions aren’t relevant for this 2025 finance app, but this avoids ambiguity and ensures invalid dates sort to the bottom.

feature/home/src/commonMain/kotlin/org/mifospay/feature/home/HomeScreen.kt
HekmatullahAmin: look if it doesn't cause any problem change all of them. it looks extra

core/common/src/commonMain/kotlin/org/mifospay/core/common/DateHelper.kt
biplab1: I added `parseDateToMillis` to handle the reverse conversion — turning a date string like `15 Aug 2025` into epoch millis. The existing helpers seem focused on formatting (millis → string), but I didn’t find one for string → millis. Please let me know if I missed an existing function that already covers this.

I updated `parseDateToMillis` to use `runCatching` with `getOrNull` as suggested. I also replaced the silent `null` return for invalid month names with an `IllegalArgumentException`, so invalid input is explicitly treated as an error while still being safely handled by `runCatching`.

core/data/src/commonMain/kotlin/org/mifospay/core/data/repositoryImpl/SelfServiceRepositoryImpl.kt
biplab1: I removed the `flowOn` (kept it earlier since others in the file still use it). Regarding removing the if/else, I tried it but it results in an empty state before loading, so I’m keeping it as is for now.

feature/history/src/commonMain/kotlin/org/mifospay/feature/history/HistoryScreen.kt
```diff
@@ -75,54 +76,61 @@ internal fun HistoryScreenContent(
...
-                is HistoryState.ViewState.Loading -> {
+        topBar = if (showTopBar && onBackClick != null) {
+            {
+                MifosTopBar(
+                    topBarTitle = stringResource(Res.string.feature_history_title),
+                    backPress = onBackClick,
+                )
+            }
+        } else {
+            {}
```
HekmatullahAmin: don't need to have else block just have a look it looks idle. use let or like below which will be empty if the condition doesn't pass.
```
topBar = {
            if (showTopBar && onBackClick != null) {
                MifosTopBar(
                    topBarTitle = "",
                    backPress = {},
                )
            }
        }
```

biplab1: RIP `else {}` 👋

PR #1928, opened 2025-09-08
feat: Comment out features not using self API

Nagarjuna0033:
This commit comments out the following features as they are not currently using the self API:

- Standing Instructions (SI)
- Invoices
- Cards
- Merchants
- KYC
- Notifications

This is a temporary measure until these features are updated to use the self API.

Jira Task: [316](https://mifosforge.jira.com/browse/MW-316)

image
image
image


HekmatullahAmin:
@Nagarjuna0033 if the goal is to remove those which are not self service then what about the `searchApi` i believe it is used in auth module during sign up for checking if the username and mobile number already exist. and also what about the `accountTransferApi` i believe that is also has been used somewhere

Nagarjuna0033:
> @Nagarjuna0033 if the goal is to remove those which are not self service then what about the `searchApi` i believe it is used in auth module during sign up for checking if the username and mobile number already exist. and also what about the `accountTransferApi` i believe that is also has been used somewhere

Yes saw now `searchApi` is not self but is that is mandatory right to check if username and mobile. I think api will throw error if username and mobile already exist

HekmatullahAmin:
> > @Nagarjuna0033 if the goal is to remove those which are not self service then what about the `searchApi` i believe it is used in auth module during sign up for checking if the username and mobile number already exist. and also what about the `accountTransferApi` i believe that is also has been used somewhere
> 
> Yes saw now `searchApi` is not self but is that is mandatory right to check if username and mobile. I think api will throw error if username and mobile already exist

@Nagarjuna0033 Yes, that’s correct our app depends on those APIs. If Mr @therajanmaurya  wants everything except self-service commented out, please list the APIs in `FineractApiManager.kt` that you haven’t commented out yet. And review that list with Mr Rajan and decide what to keep or remove.

therajanmaurya:
> > > @Nagarjuna0033 if the goal is to remove those which are not self service then what about the `searchApi` i believe it is used in auth module during sign up for checking if the username and mobile number already exist. and also what about the `accountTransferApi` i believe that is also has been used somewhere
> > 
> > 
> > Yes saw now `searchApi` is not self but is that is mandatory right to check if username and mobile. I think api will throw error if username and mobile already exist
> 
> @Nagarjuna0033 Yes, that’s correct our app depends on those APIs. If Mr @therajanmaurya wants everything except self-service commented out, please list the APIs in `FineractApiManager.kt` that you haven’t commented out yet. And review that list with Mr Rajan and decide what to keep or remove.

We only need to support and use self API.

PR #2028, opened 2026-05-12
fix(standing-instruction): reject non-positive and non-numeric amounts

bhuvan-somisetty:
The amount validation in `AddEditSIViewModel` used `any { !it.isDigit() }` to catch invalid input, but `isDigit()` returns false for `.`, so any decimal amount like `10.50` was rejected with "Amount is Invalid" even though the amount field uses `KeyboardType.Number` which allows decimal separators.

Replaced the character-level digit check with `toDoubleOrNull() == null` for format validation, and added a `<= 0` guard to reject zero and negative amounts. The null-path message is corrected to "Amount is Invalid" rather than the previous "Amount is Required".

Tested manually by entering `10.50`, `0`, `-5`, and `abc` in the standing instruction form to confirm each case is handled correctly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved amount field validation in Standing Instructions to provide clearer error messages and prevent invalid amounts (zero or negative values).

[Review Change Stack](https://app.coderabbit.ai/change-stack/openMF/mifos-pay/pull/2028)

### DIFF
--- a/feature/standing-instruction/src/commonMain/kotlin/org/mifospay/feature/standing/instruction/createOrUpdate/AddEditSIViewModel.kt
+++ b/feature/standing-instruction/src/commonMain/kotlin/org/mifospay/feature/standing/instruction/createOrUpdate/AddEditSIViewModel.kt
@@ -229,8 +229,8 @@ internal class AddEditSIViewModel(
                 data.name.isBlank() && state.isAddMode -> showError("Name is Required")
 
                 data.amount.isBlank() && state.isAddMode -> showError("Amount is Required")
-                data.amount.any { !it.isDigit() } && state.isAddMode -> showError("Amount is Invalid")
-                data.amount.toDoubleOrNull() == null && state.isAddMode -> showError("Amount is Required")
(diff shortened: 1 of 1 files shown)


niyajali:
Hi, @bhuvan-somisetty We appreciate your contribution. To facilitate seamless integration and collaboration, please ensure you have joined our Slack workspace and the relevant designated channel. We recommend thoroughly reviewing the bookmark pages, project README, and wiki to gain a comprehensive understanding of the project architecture and development guidelines. Active participation in the daily stand-up meetings is encouraged for discussions, ticket assignments, and to foster connections with the development team. Prior to requesting an assignment in the channel, please create a Jira ticket or ask for an assignment of existing tickets. All pull request descriptions must comprehensively detail the before and after changes, substantiated by relevant video or image evidence.

bhuvan-somisetty:
Thanks for the pointers - I'll make sure to join the Slack workspace and the project channel, and go through the README/wiki and Jira board before picking up further work. Rebased this PR onto latest dev in the meantime; the fix and manual test steps are described above (10.50 / 0 / -5 / abc cases). Happy to add anything else needed for review.

PR #2031, opened 2026-05-13
fix(viewmodels): replace unsafe non-null assertions with safe null handling

bhuvan-somisetty:
Three ViewModels across the app use `!!` to dereference nullable state after a null check, but because `state` is read from a `StateFlow` via a property getter, the value can change between the check and the dereference. This pattern leads to `NullPointerException` crashes that are hard to reproduce consistently.

**HomeViewModel** — `externalId!!` inside a `.filter { !it.externalId.isNullOrBlank() }` chain is technically safe but fragile; replaced with `mapNotNull` which makes the intent explicit. The `currentSelectedAccount!!` after an `if (state.currentSelectedAccount != null)` guard is a textbook case of the stale-read problem; replaced with `?.let`.

**EditProfileViewModel** — `state.profileImage!!` inside a `if (state.profileImage != null)` block reads `state` twice through the getter. Capturing it to a local `val` before the check means both uses refer to the same snapshot.

**AddEditCardViewModel** — `state.type.savedCardId!!` is used in two places when the type is `EditItem`. In `init`, replaced the `is` check with `as?` so the cast and the null check are one operation. In `initiateAddOrEditCard`, captured `state.type` as a `val` in the `when` expression so Kotlin's smart cast kicks in and the `savedCardId` property is already non-nullable.

No behavior changes — these are correctness fixes to patterns that can crash under concurrent state updates.

## Release Notes

* **Refactor**
  * Improved internal code safety and reliability in home, profile, and saved cards features through enhanced null-handling patterns and more robust state management logic.

[Review Change Stack](https://app.coderabbit.ai/change-stack/openMF/mifos-pay/pull/2031)

### DIFF
--- a/feature/home/src/commonMain/kotlin/org/mifospay/feature/home/HomeViewModel.kt
+++ b/feature/home/src/commonMain/kotlin/org/mifospay/feature/home/HomeViewModel.kt
@@ -83,8 +83,11 @@ class HomeViewModel(
 
                                 // Save account external IDs map
                                 val accountExternalIds = result.data
-                                    .filter { !it.externalId.isNullOrBlank() }
-                                    .associate { it.id to it.externalId!! }
+                                    .mapNotNull { account ->
+                                        account.externalId?.takeIf { it.isNotBlank() }
+                                            ?.let { account.id to it }
+                                    }
+                                    .toMap()
                                 preferencesRepository.updateAccountExternalIds(accountExternalIds)
 
                                 if (selected != null) {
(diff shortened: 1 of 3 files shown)


niyajali (changes requested):
The changes doesn't make any sense and invalid... Please join Slack channel if you haven't and signup Jira and check this project Jira board, join daily standup, ask for ticket assignment and work on that instead...and close this PR

Nagarjuna0033:
@therajanmaurya we can close this ticket

bhuvan-somisetty:
Understood, apologies for the noise here. I'll join the Slack workspace and Jira board, get a ticket assigned through the proper process, and only pick up work that's actually scoped and needed going forward. Thanks for the guidance.

PR #2033, opened 2026-05-14
feat(passcode): migrate mifos-authenticator libraries to v2.3.0-beta + i18n (56 locales) + resource-key cleanup

TheKalpeshPawar:
Jira Task: [MW-400](https://mifosforge.jira.com/browse/MW-400)

## Local Language Support for passcode library
I am adding screenshots for three languages only. You can check out the rest here: [mifos-passcode-cmp-PR-79](https://github.com/openMF/mifos-passcode-cmp/pull/79)

|  |  |  |  |  |  |
|---|---|---|---|---|---|
| **Urdu** (`ur`) | ur_create | ur_confirm | ur_setup | ur_setup_prompt | ur_unlock_prompt |
| **Mexican Spanish** (`es-MX`) | es-MX_create | es-MX_confirm | es-MX_setup | es-MX_setup_prompt | es-MX_unlock_prompt |
| **Hindi** (`hi`) | hi_create | hi_confirm | hi_setup | hi_setup_prompt | hi_unlock_prompt |

## Video
https://github.com/user-attachments/assets/d5096cba-4ffb-45f6-b2ed-bbf901d2ea96

## Summary

Two consecutive library upgrades plus localization, landed on the same PR:

1. **2.0.x → 2.2.x** — the deliberate decoupling release (passcode and biometrics no longer reference each other, action-bus replaced by result callbacks, biometric registration-data lifecycle moved into the biometrics library).
2. **2.2.x → 2.3.0-beta** — caller-supplied OS prompt strings and a typed `BiometricError` sealed result, so error messages and the OS biometric dialog title/subtitle/description can be localized per consumer.
3. **i18n** — 16 keys translated across **56 locales** (every locale the app already shipped, minus `en-rGB` and the legacy `iw` alias for `he`).
4. **Resource-key convention cleanup** — every passcode-module string renamed from the legacy `feature_authenticator_*` prefix to `feature_passcode_*` to match the app-wide `feature_<module>_*` convention used by `:feature:auth`, `:feature:settings`, `:feature:accounts`, etc. Dead resource keys (7 entries with no Kotlin consumers) deleted in the same pass.

The 2.2.x → 2.3.0-beta hop is a small API change layered on top of an already-substantive refactor; none of the old action-bus symbols exist on either target API.

**Final library versions:** `mifos-authenticator-passcode` **2.3.0-beta**, `mifos-authenticator-biometrics` **2.3.0-beta** (Maven Central).

### Library / version
- `gradle/libs.versions.toml`: `mifosAuthenticatorPasscode 2.0.6 → 2.3.0-beta`, `mifosAuthenticatorBiometrics 2.0.4 → 2.3.0-beta`.
- `core/data/build.gradle.kts`: now explicitly depends on the biometrics artifact (previously transitive through passcode).
- `feature/settings/build.gradle.kts`: now depends on `:feature:passcode` and the biometrics artifact directly, because the 2.3.0-beta enable-biometrics path runs through `BiometricErrorMessages` + `BiometricPromptStrings` types owned by `:feature:passcode`.
- `settings.gradle.kts`: removed a duplicate `include(":core:network")` line.
- `cmp-android/dependencies/prodReleaseRuntimeClasspath.{txt,tree.txt}`: dependency-guard baselines regenerated for the new artifacts.

### Storage adapters (split, 2.2.x)
- `MifosPasscodeAdapterImpl` trimmed to only the three passcode methods (`PasscodeStorageAdapter` no longer owns biometric blobs in 2.2.x).
- **New** `BiometricsSetupAdapterImpl` implementing `BiometricStorageAdapter` with a two-key layout (`com.mifos.biometric_registered` flag + `com.mifos.registration_data` blob).
- `RepositoryModule` binds both adapters.

### Koin + composition (2.2.x)
- `PasscodeManager(get(), MainScope()).initialize()` → `PasscodeManager(get())` (no more action-bus init).
- `PlatformAuthenticatorLocalCompositionProvider` → `PlatformAuthenticatorCompositionProvider(biometricStorageAdapter, content)` in `MifosPaySharedApp`.

### `feature/passcode` rewrite (2.2.x)
- `MifosPasscode` now uses the result-callback API: `PasscodeScreen(..., onResult, isExternalAuthEnabled, externalAuthButton)`. Added `allowBackNavigation` and `allowBiometricAuth` parameters so callers can suppress biometric bypass during change-passcode / disable-biometrics flows.
- `MifosPasscodeViewModel` rebuilt around `PasscodeResult` branches + `authProvider.isRegistered`; biometric auto-resume gated by 4 conditions (allow flag + status + step == `Enter` + isRegistered).
- `BiometricsKey`: `onAuthenticatorClick` param renamed to `onClick` to match `android-client` convention.
- `BiometricsSetupScreenViewmodel` no longer manually saves the registration blob — the library does it on `RegistrationResult.Success`.
- Nav builders updated: `internalMifosPasscodeScreen` now carries `verificationKey: String?` + `allowBiometricAuth: Boolean` for the round-trip pattern used by Settings and intra-bank transfer.

### `feature/passcode` rewrite (2.3.0-beta — caller-supplied strings + typed errors)
- **New** `feature/passcode/.../BiometricErrorMessages.kt` defining two consumer-owned data classes plus `@Composable` factories that load them from `stringResource(...)`:
  - `BiometricErrorMessages` — one string per arm of `BiometricError` (Lockout / LockoutPermanent / HardwareUnavailable / NotEnrolled / Timeout / NoSpace / InvalidRegistrationData / InvalidArguments / Unknown). Replaces the library's old `Error(message: String)` arm that bundled an English string.
  - `BiometricPromptStrings` — `title` / `subtitle` / `description` / `negativeButtonText` for the OS biometric dialog. The 2.3.0-beta library API requires the caller to pass these in.
- `MifosPasscodeViewModel.authenticateWithBiometrics(...)` now takes `errorMessages: BiometricErrorMessages` + `promptStrings: BiometricPromptStrings`, calls `provider.onAuthenticatorClick(appName, promptStrings.title, ...)`, and maps `AuthenticationResult.Error(BiometricError)` → the appropriate localized message via the `errorMessages` data class.
- `BiometricsSetupScreenViewmodel.startRegistration(...)` accepts the same two data classes and uses the same mapping for `RegistrationResult.Error`.
- `MifosPasscode.kt` resolves both data classes once at composition (via `rememberBiometricErrorMessages()` / `rememberBiometricPromptStrings()`), then ferries them across every action that crosses the VM boundary (`OnResume`, `OnAuthenticatorClick`) — `stringResource(...)` is not callable from inside a `viewModelScope.launch` block.
- `PasscodeButtonConfig` / `PasscodeSwitchConfig` field renames absorbed: `skipButtonTextStyle` removed; `switchEnabledColor → switchTrackColor`, `switchEnabledTextColor → switchUnselectedTextColor`, `switchDisabledTextColor → switchSelectedTextColor`.

### `BiometricsSetupScreen` i18n (was hardcoded English)
- The screen previously rendered three hardcoded English strings regardless of device locale: "Secure Your App", "Setup Biometrics", "Skip for Now".
- Added 3 new keys (`feature_passcode_biometric_setup_headline` / `_confirm` / `_skip`) + translations across all 56 locales, sourced from the upstream `mifos-passcode-cmp` sample app's existing translations.
- Replaced the three hardcoded `Text(...)` literals with `stringResource(...)` calls.
- **Bug fix in same screen:** the "Setup Biometrics" button used `Modifier.width(200.dp)` (fixed), which clipped longer translations. Changed to `Modifier.widthIn(min = 200.dp)` so the button grows for long copy while keeping a sensible minimum for short ones.

### Resource-key convention rename + dead-key cleanup
- All passcode-module string keys renamed `feature_authenticator_*` → `feature_passcode_*` (26 keys × 57 strings.xml files + 5 Kotlin source files). Matches the app-wide `feature_<module>_*` convention.
- 7 dead keys deleted after a code-review-graph call-site audit confirmed zero Kotlin consumers: `use_biometrics`, `fingerprint_icon`, `facescan_icon`, `keypad_icon`, `set_up_authentication_option`, `authenticate_windows_hello`, `unsupported_platform`. These were ports from the upstream sample app that mobile-wallet never wired up.
- 6 actually-used legacy strings translated across all 55 non-Spanish locales (Spanish already had them from the MW-404 union merge): 4 pulled verbatim from the upstream sample app (`ok` / `biometric_setup_description` / `biometrics_not_set_up_message` / `biometrics_not_available_message`), "Error" hardcoded per-locale, and `feature_passcode_user_not_registered_error_message` reused the same-locale translation of `feature_passcode_biometric_error_invalid_registration_data` (semantically equivalent — both surface "biometric data invalid, re-enable in settings").

### `feature/settings` (both phases)
- `ChangePasscode` → `passcodeManager.changePasscode()` (replaces removed `PasscodeAction.ChangePasscode`).
- Biometrics toggle: enable path calls `provider.registerUser(...)` directly; disable path navigates to internal passcode for verification, and the result returns via `SavedStateHandle` under `DISABLE_BIOMETRICS_VERIFICATION_KEY`, after which `authProvider.unregister()` is called.
- `SettingsViewModel.ToggleSystemAuth` action now carries `errorMessages` + `promptStrings` (2.3.0-beta contract); `SettingsScreen` resolves them via the `remember*` factories and includes them in the action.
- Biometrics registration state is now read directly from `authProvider.isRegistered` at the composable layer (removed the duplicated `isBiometricsRegistered` from `SettingsState` to avoid drift).
- `MifosNavHost` forces `allowBiometricAuth = false` for the settings flow (both change-passcode and disable-biometrics paths must not be biometric-bypassable).

### `cmp-shared` plumbing changes
- `MainUiState` → `UserState` rename (`Loading/Success` → `UnAuthenticated/Authenticated`); `MainActivity.kt` updated.
- `MifosPayViewModel.isPasscodeCreated()` → `isPasscodeNotCreated()` (inverted predicate at the single call site).
- `MifosPayViewModel.isAppUnlocked()` removed; replaced by `isAppLocked(): Boolean?` (`AppLockRepository.isAppLocked()` is now nullable — `null` means "no flag written yet", i.e. fresh install / post-logout).
- `MifosPayApp` re-auth gate tightened: pushes `reAuthMifosPasscodeScreen` only when **all four** conditions hold — backgrounded > 15 s, lock flag is non-null and `false` (= active session that was unlocked), and `passcodeManager.state.value.passcodeStep == Enter`. The `null` lock-flag branch is treated as "no session to gate".

### Localization (16 keys × 56 locales)
- 13 keys added for the 2.3.0-beta API surface in `feature/passcode/.../composeResources/values/strings.xml`: 9 `feature_passcode_biometric_error_*` (one per `BiometricError` arm) + 4 `feature_passcode_biometric_prompt_*` (title / subtitle / description / negative-button).
- 3 keys added for the `BiometricsSetupScreen` UI labels: `feature_passcode_biometric_setup_headline` / `_confirm` / `_skip`.
- **56 locale files** written, one `values-XX/strings.xml` per locale the app already supports: `af, am, ar, be, bg, bn, ca, cs, da, de, el, es, et, fa, fi, fil, fr, gu, he, hi, hr, hu, in, it, ja, kn, ko, lt, lv, mk, ml, mr, nb, nl, pa, pl, pt-rBR, pt-rPT, ro, ru, si, sk, sl, sq, sr, sv, sw, ta, te, th, tr, uk, ur, vi, zh-rCN, zh-rTW`.
- **Skipped:** `en-rGB` (identical to default), `iw` (legacy alias for `he`; Android resolves `iw` to `he` automatically).
- Pipeline: per-language-family translation review → two rounds of cross-AI review fixes (apostrophe forms in French/Afrikaans, full-width punctuation in zh-rTW, Devanagari danda in Marathi, gender-neutral Hebrew, "arguments" terminology cleanup in Tamil, etc.) → final XML generation with proper Android escaping (`'` → `\'`, `&` → `&amp;`).
- `values-es` (Spanish) is a **union merge** with the parallel MW-404 PR's Spanish translation: their 13 pre-existing UI strings + our 16 new strings. Three terminology harmonizations were applied to our half so the two blocks read consistently: `huella dactilar → huella digital`, two instances of `la biometría → los datos biométricos`. Provenance is documented inline via HTML comments.
- Verified live on Android emulator and physical device in Hindi and Arabic (RTL layout flips correctly).

### Bug fixes uncovered along the way
- **Logout half-state**: previously, an authenticated session with no passcode would silently log the user out at compose time (defensive predicate that misfired on the happy path). Fixed by routing to `LOGIN_GRAPH` deterministically when `isPasscodeNotCreated()` is true on an `Authenticated` user.
- **First-launch re-auth**: the 15-second re-auth gate was firing on first cold start with no session. Now requires `passcodeStep == Enter` so it can only push when the user genuinely has an established session.
- **401 logout loop**: `ClientRepositoryImpl.getClient()` was routed through `BaseClient` (hardcoded `mifos:password` Basic-auth fallback, no live token) instead of `SelfServiceClient`. Server returned 401 immediately after login, `GlobalAuthManager.isUnauthorized` showed the dialog, and the user was bounced back to login. Switched to `apiManager.clientsApi` so the live token is sent.
- **BiometricsSetupScreen button clipping**: see above — fixed-width button replaced with min-width that grows for long translations.

### Documentation
- KDoc added across 27 files in the active passcode/biometrics surface (cmp-shared root + nav, all of `feature/passcode`, `feature/settings` passcode/biometric paths, `feature/auth` login→passcode handoff, `feature/transfer-intrabank` auth-gate path, `core/data` adapters + repos + bindings). Mirrors `android-client` KDoc style.
- KDocs refreshed for the v2.3.0-beta surface across 20 modified files (`BiometricErrorMessages.kt`, both viewmodels, `SettingsViewModel.kt`) — explicitly call out the new `BiometricError` sealed-type mapping and the caller-supplied prompt-strings contract.

## Reviewer note: rebase conflict with MW-404

The mass `feature_authenticator_*` → `feature_passcode_*` rename will conflict on rebase with any open PR that adds new keys under the old prefix. The MW-404 Spanish translation PR is one such case — we pre-resolved its `values-es` conflict via a union merge (see provenance HTML comments in that file), but on rebase the rename will need to be re-applied to anything MW-404 adds elsewhere. Both PRs touch the same module; either order is workable, but whichever lands second will need a small sweep to update any newly-introduced key names.

[MW-400]: https://mifosforge.jira.com/browse/MW-400?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


feature/settings/src/commonMain/kotlin/org/mifospay/feature/settings/SettingsScreen.kt
```diff
@@ -56,24 +58,72 @@ import org.mifospay.core.designsystem.icon.MifosIcons
...
+    }
 
     EventsEffect(viewmodel) { event ->
         when (event) {
             SettingsEvent.OnNavigateBack -> backPress.invoke()
-            SettingsEvent.NavigateToPasscodeScreen -> navigateToPasscodeScreen()
+            SettingsEvent.NavigateToPasscodeScreen -> {
+                navigateToPasscodeScreen(
+                    if (isRegistered) DISABLE_BIOMETRICS_VERIFICATION_KEY else null,
+                )
```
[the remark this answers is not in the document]
TheKalpeshPawar: Changing the passcode doesn't require a key. It doesn't matter to it.

core/data/src/commonMain/kotlin/org/mifospay/core/data/repositoryImpl/UserVerificationRepositoryImpl.kt
biplab1: This is outside the scope of this PR, however, just out of curiosity - Is it safe to use `Clock.System.now()` to calculate time elapsed? It could be manipulated by changing wall clock time.

biplab1 (dismissed):
Looks good to me. This can be merged.

biplab1:
@TheKalpeshPawar Please resolve the comments by CodeRabbitAI so it can be merged.

core/data/src/commonMain/kotlin/org/mifospay/core/data/repositoryImpl/UserVerificationRepositoryImpl.kt
TheKalpeshPawar: fixed

niyajali (approved):
@TheKalpeshPawar This PR implement/fixes multiple tasks though I'm merging it now but in future spilt into multiple PRs to review properly.

PR #2037, opened 2026-05-15
fix(savedcards): refresh card list after delete and wire pull-to-refresh

bhuvan-somisetty:
Deleting a saved card left the list showing the deleted entry until you navigated away and back. The root cause was that `cardState` was built directly from a one-shot API flow: once it emitted and completed, there was nothing to trigger a re-fetch after a delete.

The fix introduces a `MutableSharedFlow<Unit>` as a refresh trigger (seeded with an initial emission so the first load still happens automatically). The `cardState` flow now uses `flatMapLatest` on that trigger, so every `tryEmit(Unit)` restarts the underlying `getSavedCards` call and the list updates immediately.

Two places now emit to the trigger:
- After a successful delete, so the removed card disappears right away
- When `CardAction.RefreshCards` is dispatched, which the screen sends on pull-to-refresh

The `isPullRefreshEnabled` property was already defined on `ViewState` but the scaffold was never wired up to it. This PR connects it via `rememberMifosPullToRefreshState`, so users can also manually refresh the list by pulling down.

The known-issue comment on `CardsScreen` is removed since the underlying problem is resolved.

Tested manually: deleting a card now removes it from the list without any navigation, and pull-to-refresh re-fetches correctly.

## Summary by CodeRabbit

* **New Features**
  * Pull-to-refresh support added to saved cards screen—swipe down to manually refresh your card list.
  * Saved cards automatically refresh after successful deletion.

[Review Change Stack](https://app.coderabbit.ai/change-stack/openMF/mifos-pay/pull/2037)

### DIFF
--- a/feature/savedcards/src/commonMain/kotlin/org/mifospay/feature/savedcards/CardsScreen.kt
+++ b/feature/savedcards/src/commonMain/kotlin/org/mifospay/feature/savedcards/CardsScreen.kt
@@ -56,6 +56,7 @@ import org.mifospay.core.designsystem.component.LoadingDialogState
 import org.mifospay.core.designsystem.component.MifosBasicDialog
 import org.mifospay.core.designsystem.component.MifosLoadingDialog
 import org.mifospay.core.designsystem.component.MifosScaffold
+import org.mifospay.core.designsystem.component.rememberMifosPullToRefreshState
 import org.mifospay.core.designsystem.icon.MifosIcons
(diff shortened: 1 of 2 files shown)


bhuvan-somisetty:
Rebased on latest dev and pushed a fix for the CodeRabbit finding: both refresh call sites now go through a `requestRefresh()` helper that uses the suspending `emit` instead of `tryEmit`, so a refresh event can no longer be silently dropped under backpressure.

PR #2054, opened 2026-07-18
Feat: Pocket Repository

Kartikey15dem:
Fixes - [Jira-#MW-384](https://mifosforge.jira.com/browse/MW-384)

## Summary
This PR introduces the `PocketRepositoryImp`, establishing the core data layer for the new Pocket feature. The implementation utilizes a **Single Source of Truth (SSOT)** caching architecture to ensure seamless synchronization between screens, while employing highly optimized network logic to minimize API calls and prevent UI lag.

---

### 1. Unified Cache (Single Source of Truth)
Both the Pocket Dashboard and Manage Pockets screens rely on a single `detailedPocketCache` (`MutableStateFlow`). This is populated by merging base pockets with the account details, eliminating redundant API calls and state mismatch.

### 2. Share Account Balance Resolution & Safe Fallbacks
Unlike Loan and Savings accounts, the base `getClientAccounts` response omits the data needed to calculate Share account balances. 
* We explicitly call `getShareAccountDetails` to calculate the true monetary balance (`totalApprovedShares * currentMarketPrice`). 
* **Safe Fallback:** This is wrapped in a `try/catch` block so that if the specific share API fails, the app gracefully falls back to displaying the product name with a blank balance rather than crashing the flow.

### 3. Optimized Linking (The Mapping ID Strategy)
We completely bypass heavy, full-profile resyncs when a user links an account.
* **Model Reuse:** `linkAccounts` directly receives the `explicitlyAddedAccount` model from the UI, reusing the heavy details (like live share prices) already calculated during the Bottom Sheet fetch.
* **The Double-Fetch:** After the link API succeeds, we fetch *only* the basic pockets from the network. This lightweight call grabs the newly generated backend **Mapping ID** (required for future delinking) and attaches it to the reused model before injecting it into the cache.

### 4. Surgical Delinking
Delinking hits the backend and instantly filters that Mapping ID out of the local cache. This provides an immediate, zero-latency UI update without any network refetching.

## Summary by CodeRabbit

* **New Features**
  * Added support for pocket accounts spanning savings, loan, and share account types.
  * Added detailed pocket account data, including balances, currencies, statuses, and lifecycle states.
  * Added account linking and delinking.
  * Added discovery of linkable accounts for pocket linking.
  * Enhanced share-account balance enrichment when linking/opening detailed pockets.
  * Introduced consistent loading/success/error handling with offline-aware network checks.
* **Model & API Updates**
  * Updated network DTOs to better handle optional/nullable fields across pocket and share data.

### DIFF
--- a/core/network/src/commonMain/kotlin/org/mifospay/core/network/model/entity/loanAccount/LoanSummaryResponseDto.kt
+++ b/core/network/src/commonMain/kotlin/org/mifospay/core/network/model/entity/loanAccount/LoanSummaryResponseDto.kt
@@ -38,7 +38,7 @@ data class LoanSummaryResponseDto(
 
     val totalOutstanding: Double = 0.0,
 
-    private val overdueSinceDate: List<Int>? = null,
+    val overdueSinceDate: List<Int>? = null,
 
     val currency: CurrencyResponseDto? = null,
 )
--- a/core/network/src/commonMain/kotlin/org/mifospay/core/network/model/entity/loanAccount/LoanSummaryResponseDto.kt
+++ b/core/network/src/commonMain/kotlin/org/mifospay/core/network/model/entity/loanAccount/LoanSummaryResponseDto.kt
@@ -14,29 +14,29 @@ import org.mifospay.core.network.model.entity.common.CurrencyResponseDto
 
 @Serializable
 data class LoanSummaryResponseDto(
-    val principalDisbursed: Double = 0.0,
+    val principalDisbursed: Double? = null,
 
-    val principalPaid: Double = 0.0,
+    val principalPaid: Double? = null,
 
-    val interestCharged: Double = 0.0,
+    val interestCharged: Double? = null,
 
-    val interestPaid: Double = 0.0,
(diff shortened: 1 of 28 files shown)


core/network/src/commonMain/kotlin/org/mifospay/core/network/model/entity/loanAccount/LoanSummaryResponseDto.kt
```diff
@@ -0,0 +1,44 @@
...
+
+    val totalExpectedRepayment: Double = 0.0,
+
+    val interestWaived: Double = 0.0,
+
+    val totalRepayment: Double = 0.0,
+
+    val feeChargesWaived: Double = 0.0,
+
+    val totalOutstanding: Double = 0.0,
```
Nagarjuna0033: @Kartikey15dem make all these fields as nullable

Nagarjuna0033 (changes requested):

Nagarjuna0033 (approved):

revanthkumarJ (approved):

